PR #5, opened 2019-01-08
Add ;invite and shutdown message (console) also add readme

squeezeboxyt:
# Pull request

 - [x] I have read and agreed to the [CODE OF CONDUCT](https://github.com/CascadeBot/CascadeBot/blob/master/CODE_OF_CONDUCT.md).
 - [x] What I'm implementing was assigned to me or was not being worked on and is a somewhat requested feature. For reference see our [trello](https://trello.com/b/gUQehy4l/bot).
 - [ ] I have tested all of my changes.
 - [x] I have followed the style of the rest of the project.
 
## Added/Changed
Added InviteCommand.java, added a shutdown message using Logger and also changed the spacing in a comment in CascadeBot.java because it was really annoying me. Also added a README.md

## Feature
We should add ;invite to the bot because it gives users a chance to invite the bot to their own server, and spread the love 'n good vibes.

PR #6, opened 2019-01-08
Mongo pojo

binaryoverload:
- No more migrations in-place on the bot, we'll use a separate program for that
- Rename `CascadeBot.instance()` to `CascadeBot.INS` which is a lot easier to use


binaryoverload:
Waiting on #5

PR #15, opened 2019-01-28
Global Emotes

binaryoverload:
# Pull request

 - [x] I have read and agreed to the [CODE OF CONDUCT](https://github.com/CascadeBot/CascadeBot/blob/master/CODE_OF_CONDUCT.md).
 - [x] What I'm implementing was assigned to me or was not being worked on and is a somewhat requested feature. For reference see our [trello](https://trello.com/b/gUQehy4l/bot).
 - [x] I have followed the style of the rest of the project.
 
## Added/Changed
- Added reading global emotes from config
- Added helper method for easy display of emotes


config.example.yml
```diff
@@ -62,6 +62,19 @@ security_levels:
   developer:
   staff:
 
+
+# Global emotes the bot can use to display various information
+global_emotes:
```
weeryan17: We should make this a map. i feel like hardcoding this is a bad idea. Like to boils down to a map anyway

src/main/java/com/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
```diff
@@ -54,6 +54,11 @@ public CommandType getType() {
         return CommandType.INFORMATIONAL;
     }
 
+    @Override
+    public Permission getPermission() {
+        return Permission.SERVER_INFO_COMMAND;
```
weeryan17: It begins, but yegh why is permissions stuff in this pr

src/main/java/com/cascadebot/cascadebot/data/objects/GuildPermissions.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2019 CascadeBot. All rights reserved.
+ * Licensed under the MIT license.
+ */
+
+package com.cascadebot.cascadebot.data.objects;
+
+public class GuildPermissions {
```
weeryan17: See previous review

src/main/java/com/cascadebot/cascadebot/permissions/Permission.java
```diff
@@ -0,0 +1,105 @@
...
+package com.cascadebot.cascadebot.permissions;
+
+import com.cascadebot.cascadebot.commandmeta.CommandType;
+import com.cascadebot.cascadebot.utils.CollectionUtils;
+
+import java.util.Arrays;
+import java.util.EnumSet;
+import java.util.Map;
+
+public enum Permission {
```
weeryan17: Do i even need to say it

src/main/java/com/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
binaryoverload: Oh 🤔

config.example.yml
binaryoverload: Why is it a bad idea? The IDs litterally never change

binaryoverload:
@weeryan17 Permissions is stuck in this for now, I haven't really implemented anything to do with handling permissions so it's fine

config.example.yml
weeryan17: I meant it's a bad idea to hardcode the names, not the ids.

binaryoverload: Oh the names aren't hardcoded, they can be literally anything

binaryoverload:
Yeah sorry about that, this is the reason I'm transitioning to branches for stuff like this

PR #17, opened 2019-01-31
Update issue templates

binaryoverload:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "[Bug] "
+labels: bug
+assignees: ''
```
weeryan17: why exactly is section this here?

binaryoverload: It's a thing that GitHub adds, these are basically presets for issues that users can select from a list

weeryan17: fair enough, just never seen them before so.

PR #20, opened 2019-02-08
Add sentry to the bot for reporting errors

binaryoverload:
Self explanatory really

### DIFF
--- a/config.example.yml
+++ b/config.example.yml
@@ -19,6 +19,9 @@
 # Whether to output DEBUG messages to STDOUT
 debug: false
 
+# The Sentry DSN to use for reporting errors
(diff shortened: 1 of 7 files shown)


weeryan17:
have you tested this?

binaryoverload:
There's no way I can test it currently but this is the exact way we did it on FlareBot so it'll work

weeryan17:
We should set up a sentry account and test before we merge this, cause i know sentry has changed some of their stuff.

binaryoverload:
I'll do that this evening then 👌

binaryoverload:
@weeryan17 Fully tested, changed it to use an env variable as it works better with the logback appender.

PR #22, opened 2019-02-09
Add documentation and code standards

binaryoverload:


binaryoverload:
@mrjvs I know you're not a java dev but if you could look over the templates/guidelines it'd be appreciated :D

mrjvs:
Looks solid. No complaints.

src/main/java/com/cascadebot/cascadebot/messaging/MessageType.java
```diff
@@ -7,15 +7,15 @@
...
+import java.awt.Color;
 
 public enum MessageType {
 
-    INFO(Color.CYAN, "\u2139"), // ℹ
-    SUCCESS(Color.GREEN, "\u2705"), // ✅
-    WARNING(Color.YELLOW, "\u26A0"), // ⚠
-    MODERATION(Color.WHITE, "\uD83D\uDC6E"), // 👮
-    DANGER(Color.RED, "\u274C"), // ❌
+    INFO(Color.CYAN, "\u2139"), // Information icon
```
weeryan17: I feel like we should also include the unicode character

src/main/java/com/cascadebot/cascadebot/utils/FormatUtils.java
```diff
@@ -128,28 +128,28 @@ private static String getFormattedInlineFields(List<MessageEmbed.Field> fieldLis
...
 
         sb.append('`');
 
-        for(String head : header) {
+        for (String head : header) {
             sb.append(String.format("%-25s", head));
         }
 
-        sb.append("\u200B`");
+        sb.append("\u200B`"); // Zero width space
```
weeryan17: In reference to my previous statement having this just be text is fine as zero width space has no content

binaryoverload: Yeah but immediately looking at this you don't necessarily know it's a zero width space hence the comment

src/main/java/com/cascadebot/cascadebot/messaging/MessageType.java
binaryoverload: Okay

src/main/java/com/cascadebot/cascadebot/utils/FormatUtils.java
weeryan17: Yegh having the text is fine. i was more referring to how we can't really put the unicode character here as zero with space is well 0 width

weeryan17: So I was saying ignore my previous review comment in this case

binaryoverload: Ah sorry for the misunderstanding, yeah I agree

binaryoverload:
@weeryan17 have you looked over the guideline stuff (Especially the code style document)?

weeryan17:
@binaryoverload yep

PR #26, opened 2019-02-14
Add module functionality

binaryoverload:


binaryoverload:
@weeryan17 All sets from the guild data that have methods to add/remove from them (Enabled commands and modules) are now immutable to stop invalid modification

src/main/java/com/cascadebot/cascadebot/events/CommandListener.java
```diff
@@ -60,6 +64,23 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
...
+            if (cmd.getModule().isPublicModule() &&
+                    !guildData.isModuleEnabled(cmd.getModule())) {
+                if (guildData.willDisplayModuleErrors()) {
+                    EmbedBuilder builder = MessagingObjects.getClearThreadLocalEmbedBuilder();
+                    builder.setDescription(String.format("Module `%s` disabled!", cmd.getModule().toString()));
+                    builder.setTimestamp(Instant.now());
+                    builder.setFooter("Requested by " + event.getAuthor().getAsTag(), event.getAuthor().getEffectiveAvatarUrl());
+                    Messaging.sendDangerMessage(event.getChannel(), builder, guildData.getUseEmbedForMessages());
+                }
+                CascadeBot.logger.info(
```
weeryan17: We should maybe add this to the modlog when we work on that but considering modules are a per guild thing I don't think logging to the console is the best

src/main/java/com/cascadebot/cascadebot/permissions/Permission.java
```diff
@@ -5,25 +5,27 @@
 
 package com.cascadebot.cascadebot.permissions;
 
-import com.cascadebot.cascadebot.commandmeta.CommandType;
+import com.cascadebot.cascadebot.commandmeta.Module;
 
 import java.util.Arrays;
 import java.util.EnumSet;
 
 public class Permission {
```
weeryan17: We might want to rename our permission to like "CascadPermission" so we don't have problmes with jda permissions.

src/main/java/com/cascadebot/cascadebot/events/CommandListener.java
binaryoverload: Hmm, my thoughts were that it's bound to be less spammy than command logging that we already do. I'll remove it

weeryan17: I mean it's spam for a thing that does go through. commands actually do something.

PR #43, opened 2019-02-20
Core/usage

weeryan17:
I want to get feedback on the usage system before we merge it so I'm opening this pr.
(this will also give binary a chance to improve my terrible descriptions)

I still need to implement:

- [x] aliases
- [x] Utils so we can use the arguments for more then just usage

### DIFF
--- a/src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
+++ b/src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
@@ -56,7 +56,7 @@ public class Argument {
      * Gets the usage string.
      *
      * Formatting:
-     *  - Aliased arguments are shown as {@code <alias1/alias2}} for as many aliases as the argument has.
+     *  - Aliased arguments are shown as {@code <alias1/alias2>} for as many aliases as the argument has.
      *  - A required parameter is show as {@code <argument>}
      *  - An optional parameter is show as {@code [argument]}
      *
--- a/src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
+++ b/src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
@@ -29,39 +29,39 @@ public class Argument {
     }
 
     public static Argument of(String arg) {
-        return new Argument(arg, "", new HashSet<>(), ArgumentType.COMMAND, new HashSet<>());
+        return new Argument(arg, "", Set.of(), ArgumentType.COMMAND, Set.of());
     }
 
     public static Argument ofA(String arg, Set<String> aliases) {
-        return new Argument(arg, "", new HashSet<>(), ArgumentType.COMMAND, aliases);
+        return new Argument(arg, "", Set.of(), ArgumentType.COMMAND, aliases);
     }
 
     public static Argument of(String arg, ArgumentType type) {
-        return new Argument(arg, "", new HashSet<>(), type, new HashSet<>());
+        return new Argument(arg, "", Set.of(), type, Set.of());
     }
 
     public static Argument ofA(String arg, ArgumentType type, Set<String> aliases) {
-        return new Argument(arg, "", new HashSet<>(), type, aliases);
+        return new Argument(arg, "", Set.of(), type, aliases);
     }
 
     public static Argument of(String arg, String description) {
-        return new Argument(arg, description, new HashSet<>(), ArgumentType.COMMAND, new HashSet<>());
+        return new Argument(arg, description, Set.of(), ArgumentType.COMMAND, Set.of());
     }
 
     public static Argument ofA(String arg, String description, Set<String> aliases) {
-        return new Argument(arg, description, new HashSet<>(), ArgumentType.COMMAND, aliases);
+        return new Argument(arg, description, Set.of(), ArgumentType.COMMAND, aliases);
     }
 
     public static Argument of(String arg, String description, ArgumentType type) {
-        return new Argument(arg, description, new HashSet<>(), type, new HashSet<>());
+        return new Argument(arg, description, Set.of(), type, Set.of());
     }
 
     public static Argument ofA(String arg, String description, ArgumentType type, Set<String> aliases) {
-        return new Argument(arg, description, new HashSet<>(), type, aliases);
+        return new Argument(arg, description, Set.of(), type, aliases);
     }
 
     public static Argument of(String arg, String description, Set<Argument> subArgs) {
-        return new Argument(arg, description, subArgs, ArgumentType.COMMAND, new HashSet<>());
+        return new Argument(arg, description, subArgs, ArgumentType.COMMAND, Set.of());
     }
 
     public static Argument ofA(String arg, String description, Set<Argument> subArgs, Set<String> aliases) {
--- a/src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
+++ b/src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
@@ -28,22 +28,6 @@ public class Argument {
         this.aliases = Collections.unmodifiableSet(aliases);
     }
 
-    public static Argument of(String arg) {
-        return new Argument(arg, "", Set.of(), ArgumentType.COMMAND, Set.of());
-    }
-
-    public static Argument ofA(String arg, Set<String> aliases) {
-        return new Argument(arg, "", Set.of(), ArgumentType.COMMAND, aliases);
-    }
-
-    public static Argument of(String arg, ArgumentType type) {
-        return new Argument(arg, "", Set.of(), type, Set.of());
-    }
-
-    public static Argument ofA(String arg, ArgumentType type, Set<String> aliases) {
-        return new Argument(arg, "", Set.of(), type, aliases);
-    }
-
     public static Argument of(String arg, String description) {
         return new Argument(arg, description, Set.of(), ArgumentType.COMMAND, Set.of());
     }
--- a/src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
+++ b/src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
@@ -46,14 +46,14 @@ public class Argument {
         return new Argument(arg, description, subArgs, ArgumentType.COMMAND);
     }
 
-    protected String getUnformattedUsageString(String command) {
+    protected String getUnformattedUsageString(String base) {
         StringBuilder usageBuilder = new StringBuilder();
         if (subArgs.size() > 0) {
             if (!description.isBlank()) {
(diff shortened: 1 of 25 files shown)


binaryoverload:
Instead of having a set of subarguments, maybe each argument had a child argument and a parent argument. This would allow preserving the order of arguments in a more consistent manner and mean that things follow a logical structure. This would be harder to construct but a simple helper method would make it no more difficult than current.

Just a thought.

weeryan17:
Well they'd have to not have a child argument as that would break allowing multiple sub args, but we could do parent, but i feel like that's harder to manage. I mean if we wanted to preserve order we can use lists witch would be a lot simpler to do.

src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+public class Argument {
+
+    private String arg;
+    private String description;
+    private Set<Argument> subArgs;
+    private ArgumentType type;
+    private Set<String> aliases;
```
binaryoverload: Since this will effectively be entirely immutable, every field should be set to final

src/main/java/com/cascadebot/cascadebot/commandmeta/ICommandExecutable.java
```diff
@@ -8,6 +8,9 @@
 import com.cascadebot.cascadebot.permissions.CascadePermission;
 import net.dv8tion.jda.core.entities.Member;
 
+import java.util.HashSet;
```
binaryoverload: Unused import

src/main/java/com/cascadebot/cascadebot/commandmeta/ICommandExecutable.java
```diff
@@ -16,8 +19,19 @@
...
     CascadePermission getPermission();
 
+    String description();
+
     default boolean deleteMessages() {
         return true;
     }
 
+    /**
+     * Not when overriding this. the system automatically handles sub command arguments so you only need to define arguments here that are not defined their.
```
binaryoverload: `not when overriding this`?

src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public static Argument ofA(String arg, String description, ArgumentType type, Set<String> aliases) {
+        return new Argument(arg, description, Set.of(), type, aliases);
+    }
+
+    public static Argument of(String arg, String description, Set<Argument> subArgs) {
+        return new Argument(arg, description, subArgs, ArgumentType.COMMAND, Set.of());
+    }
+
+    public static Argument ofA(String arg, String description, Set<Argument> subArgs, Set<String> aliases) {
```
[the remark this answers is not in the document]
binaryoverload: Also `ofA` is not a great name, why not just stick with an overloaded `of`?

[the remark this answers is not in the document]
weeryan17: Because java considers sets even with different types to be the same thing method wise.
Also what factory methods should i remove?

src/main/java/com/cascadebot/cascadebot/commandmeta/ICommandExecutable.java
weeryan17: whops my bad

binaryoverload: What does it mean 👀

src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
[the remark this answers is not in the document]
binaryoverload: I think the ones without a description. In my mind, all arguments should have a description. If we don't want a description we can pass in `""` but I don't think it'll be that often

src/main/java/com/cascadebot/cascadebot/commandmeta/ICommandExecutable.java
weeryan17: Actually what is says is correct. I wanted to specify that sub commands are automatically handled so you don't need this method to handle arguments for sub commands (at least at the top sub command level)

src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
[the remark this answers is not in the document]
weeryan17: fair enough

src/main/java/com/cascadebot/cascadebot/commandmeta/ArgumentType.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) 2019 CascadeBot. All rights reserved.
+ * Licensed under the MIT license.
+ */
+
+package com.cascadebot.cascadebot.commandmeta;
+
+public enum ArgumentType {
+    /**
+     * Represents an option param
```
binaryoverload:
```suggestion
     * Represents an optional parameter.
```

src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
```diff
@@ -0,0 +1,150 @@
...
+                usageBuilder.append(" - ").append(description);
+            }
+            usageBuilder.append('\n');
+        }
+
+        return usageBuilder.toString();
+    }
+
+    /**
+     * Checks for this argument at a giving position.
```
binaryoverload:
```suggestion
     * Checks for this argument at a given position.
```

src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
```diff
@@ -0,0 +1,150 @@
...
+
+    public static Argument ofA(String arg, String description, Set<Argument> subArgs, Set<String> aliases) {
+        return new Argument(arg, description, subArgs, ArgumentType.COMMAND, aliases);
+    }
+
+    /**
+     * Gets the usage string.
+     *
+     * @param base The base command/preifx to use. Example: ';help '
+     * @return A string representing the usage
```
binaryoverload: Full stops at end of lines please

src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
```diff
@@ -0,0 +1,150 @@
...
+
+        return usageBuilder.toString();
+    }
+
+    /**
+     * Checks for this argument at a giving position.
+     *
+     * @param args The args sent in from the command
+     * @param pos  The position this argument should be in
+     * @return If the argument exists at that position
```
binaryoverload: Full stops at end of lines please

src/main/java/com/cascadebot/cascadebot/commandmeta/ArgumentType.java
```diff
@@ -0,0 +1,21 @@
...
+     */
+    OPTIONAL,
+    /**
+     * Represents an required param
```
binaryoverload:
```suggestion
     * Represents an required parameter.
```

src/main/java/com/cascadebot/cascadebot/commandmeta/ArgumentType.java
```diff
@@ -0,0 +1,21 @@
...
+     */
+    REQUIRED,
+    /**
+     * Represents that this is just part of the command and isn't a pram
```
binaryoverload:
```suggestion
     * Represents that this is just part of the command and isn't a parameter.
```

src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
binaryoverload: `args` should be `arguments` in docs

src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    public static Argument ofA(String arg, String description, Set<Argument> subArgs, Set<String> aliases) {
+        return new Argument(arg, description, subArgs, ArgumentType.COMMAND, aliases);
+    }
+
+    /**
+     * Gets the usage string.
+     *
+     * @param base The base command/preifx to use. Example: ';help '
```
binaryoverload:
```suggestion
     * @param base The base command/prefix to use. Example: ';help '
```

src/main/java/com/cascadebot/cascadebot/commandmeta/Argument.java
```diff
@@ -0,0 +1,150 @@
...
+    public static Argument of(String arg, String description, Set<Argument> subArgs) {
+        return new Argument(arg, description, subArgs, ArgumentType.COMMAND, Set.of());
+    }
+
+    public static Argument ofA(String arg, String description, Set<Argument> subArgs, Set<String> aliases) {
+        return new Argument(arg, description, subArgs, ArgumentType.COMMAND, aliases);
+    }
+
+    /**
+     * Gets the usage string.
```
binaryoverload: Could you give an example of how each argument is represented as well as an example showing how a line would be constructed?

weeryan17: Not really without having a 50 line documentation as it's really different depending on what the args is if it has sub args if it has a description. if you can come up with a way to do so feel free to put it in here.

src/main/java/com/cascadebot/cascadebot/commandmeta/CommandContext.java
```diff
@@ -248,6 +250,42 @@ public MessageAction replyImage(String url) {
         }
     }
 
+    public void replyUsage(ICommandExecutable command) {
```
binaryoverload: Would this be better to have a `String` as a parameter? This would allow to get a usage easily from elsewhere other than in the command.

weeryan17: We could add a string method, but we should still have reply usage as a fast way to send it

weeryan17: I don't see the need

binaryoverload:
I don't think I explained myself properly, I was talking more about having a replyUsage(String command) method. Having a getUsage command that returned a string wouldn't be a bad idea but I don't think we'd use it

weeryan17:
@binaryoverload yegh i got that after i made the getusage method. I don't see the need for getting the usage outside of a command

binaryoverload:
Yeah true, are you going to revert that change?

weeryan17:
probably not. not harming anything and we might use it in the future

PR #59, opened 2019-03-13
Command/Feature flagging

### LINKED ISSUE #58: Command/Feature flagging
Add functionality to add flags to things so we can check whether this feature should be done. For now, this will just be the `beta` flag to introduce commands/features that are only for the beta program.

The flags I can forsee us needing:
 - Alpha??
 - Beta
 - Patreon Tier 1    
 - Patreon Tier 2   
 - Patreon Tier 3    

**These are just examples for the Patreon!**
I won't be implementing anything other than the first 2 for now as we don't know how we will structure our Patreon stuff.

---

Implementation proposal:
 - `Flags.java` enum to contain the flags
 - Introduce a `getFlags()` method to `ICommandExecutable` that returns an EnumSet representing the flags it requires to execute
 - Add a Set of enabled flags into guild data
    - This will be used to enable things like beta or Patreon access
 - Add a subcommand to `guild` to enable and disable beta
   - `;guild beta [guildId]`
   - Maybe this could be modified to work with all flags in the enum?

binaryoverload:
Closes #58

### DIFF
--- a/src/main/java/com/cascadebot/cascadebot/commandmeta/ICommandExecutable.java
+++ b/src/main/java/com/cascadebot/cascadebot/commandmeta/ICommandExecutable.java
@@ -5,7 +5,7 @@
 
 package com.cascadebot.cascadebot.commandmeta;
 
-import com.cascadebot.cascadebot.data.objects.Flags;
+import com.cascadebot.cascadebot.data.objects.Flag;
 import com.cascadebot.cascadebot.permissions.CascadePermission;
 import net.dv8tion.jda.core.entities.Member;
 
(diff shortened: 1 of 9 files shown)


binaryoverload:
@mrjvs @weeryan17 Question, should I continue doing what I'm doing with not having a flag subcommand and just having this:
`;guild [flag] [guildId] - toggles a flag`
`;guild list-flags - lists flags`
or have a flag subcommand like this
`;guild flag [flag] [guildId] - toggles a flag`
`;guild flag list - lists flags`

It doesn't particularly matter as it's a dev-only command just looking for some feedback cause I'm a perfectionist 😅

binaryoverload:
Or any @CascadeBot/contributors for that matter

Ikinon:
The way you are doing it now is fine.

weeryan17:
Make it a sub command to allow for future expansion.

PR #60, opened 2019-03-17
Add hi and goodbye

binaryoverload:
# Pull request

So I think we should have hello and goodbye messages that are sent when we are added/removed from a guild. This will allow us to include things like: 
 - Links to our panel
 - A removal survey
 - Awareness on things people need to be aware of
 - Etc etc

### DIFF
--- a/config.example.yml
+++ b/config.example.yml
@@ -19,6 +19,17 @@
 # Whether to output DEBUG messages to STDOUT
 debug: false
 
+# This is the message that will either be DMed to the guild owner
+# or posted in the first text channel whichever comes first.
+guild_welcome: |-
+  Thank you for adding CascadeBot!
+
+# This is the message that will be DMed to the guild owner
+# when we are removed from a guild.
+guild_goodbye: |-
+  We're sorry that this is goodbye! We hope we see you again soon :(
+  If you have a moment, please let us know what we did wrong here: <insert form link>
+
 # REQUIRED
 # The token and client ID for the bot, kinda self explanatory...
 bot:
--- a/src/main/java/com/cascadebot/cascadebot/events/GeneralEvents.java
+++ b/src/main/java/com/cascadebot/cascadebot/events/GeneralEvents.java
@@ -31,11 +31,7 @@ public class GeneralEvents extends ListenerAdapter {
             Guild guild = event.getGuild();
             guild.getOwner().getUser().openPrivateChannel().queue(privateChannel -> {
                 privateChannel.sendMessage(Config.INS.getGuildWelcomeMessage()).queue();
-            }, error -> {
-                if (guild.getDefaultChannel() != null) {
-                    guild.getDefaultChannel().sendMessage(Config.INS.getGuildWelcomeMessage()).queue();
-                }
-            });
+            }, error -> { /* Do nothing */ });
         }
     }
 
@@ -44,7 +40,7 @@ public class GeneralEvents extends ListenerAdapter {
         if (!StringUtils.isBlank(Config.INS.getGuildGoodbyeMessage())) {
             event.getGuild().getOwner().getUser().openPrivateChannel().queue(privateChannel -> {
                 privateChannel.sendMessage(Config.INS.getGuildGoodbyeMessage()).queue();
-            });
+            }, error -> { /* Do nothing */ });
         }
     }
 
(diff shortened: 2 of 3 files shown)


config.example.yml
```diff
@@ -19,6 +19,17 @@
...
+# This is the message that will either be DMed to the guild owner
+# or posted in the first text channel whichever comes first.
+guild_welcome: |-
+  Thank you for adding CascadeBot!
+
+# This is the message that will be DMed to the guild owner
+# when we are removed from a guild.
+guild_goodbye: |-
+  We're sorry that this is goodbye! We hope we see you again soon :(
+  If you have a moment, please let us know what we did wrong here: <insert form link>
```
JamesHawkinss: I suggest changing your wording here. We might not have done something wrong, perhaps say something like "please let us know why this is goodbye here:"

binaryoverload:
This is just an example, we'll come up with a lot better one in future. And imo we should always act if we have done something wrong from a marketing angle (Because if we've lost a server, we've failed 😂 )

src/main/java/com/cascadebot/cascadebot/events/GeneralEvents.java
```diff
@@ -20,4 +25,27 @@ public void onReady(ReadyEvent event) {
...
     }
 
+    @Override
+    public void onGuildJoin(GuildJoinEvent event) {
+        if (!StringUtils.isBlank(Config.INS.getGuildWelcomeMessage())) {
+            Guild guild = event.getGuild();
+            guild.getOwner().getUser().openPrivateChannel().queue(privateChannel -> {
+                privateChannel.sendMessage(Config.INS.getGuildWelcomeMessage()).queue();
+            }, error -> {
+                if (guild.getDefaultChannel() != null) {
```
weeryan17: Aren't default channels not a thing any more? also we should send in a channel at all if the owners dms are closed imo.

mrjvs:
Im against the idea of using DM's to convey these types of things. if people leave, so be it. they dont want to be pestered by surveys or messages. they just want to leave since they are done investigating the server. having a welcome message in a text channel to show the rules and welcome them is something I like tho.

src/main/java/com/cascadebot/cascadebot/events/GeneralEvents.java
binaryoverload: In JDA a default channel is defined as:
```
This channel is the first channel in the guild (ordered by position) that the getPublicRole() has the Permission.MESSAGE_READ in.
```
Good point about whether we should or not 🤔

JamesHawkinss:
> Im against the idea of using DM's to convey these types of things. if people leave, so be it. they dont want to be pestered by surveys or messages. they just want to leave since they are done investigating the server. having a welcome message in a text channel to show the rules and welcome them is something I like tho.

This is for when the bot is removed or added from a guild, not when a person leaves or joins

mrjvs:
> 
> 
> > Im against the idea of using DM's to convey these types of things. if people leave, so be it. they dont want to be pestered by surveys or messages. they just want to leave since they are done investigating the server. having a welcome message in a text channel to show the rules and welcome them is something I like tho.
> 
> This is for when the bot is removed or added from a guild, not when a person leaves or joins

erhm, shush. I knew that 😓

binaryoverload:
@mrjvs Opinion on sending in the first available channel if DMs not available?

mrjvs:
> 
> 
> @mrjvs Opinion on sending in the first available channel if DMs not available?

dont, dm's are disabled for a reason. and on big servers you dont want bots to randomly send messages.

JamesHawkinss:
> @mrjvs Opinion on sending in the first available channel if DMs not available?

@binaryoverload I know I wasn't asked but I think it should just DM, not send to first available channel

Ikinon:
@binaryoverload
I know I was not involved but:
What if we searched for a channel named something like "bot-channel"
If channel don't exist just leave it?

JamesHawkinss:
> @binaryoverload
> I know I was not involved but:
> What if we searched for a channel named something like "bot-channel"
> If channel don't exist just leave it?

We would have to do a lot of research to effectively put this in place, because loads of people call it loads of different things

Ikinon:
True

binaryoverload:
> @binaryoverload
> I know I was not involved but:
> What if we searched for a channel named something like "bot-channel"
> If channel don't exist just leave it?

We could do that but like @JamesHawkinss said people call them loads of different things. I'll just remove it :D

PR #73, opened 2019-03-27
Suggestion: Move player storage into GuildData

binaryoverload:
I think maybe keeping it in GuildData allows for more direct storage of the player

🤷‍♂️ let me know what you think


JamesHawkinss:
good idea! +1

PR #75, opened 2019-03-28
Move PlayerListener

binaryoverload:
As discussed I've moved `PlayerListener` into the `events` package with everything else and named `LoopType` to `LoopMode` as well as including more conventional getters and setters

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
+++ b/src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
@@ -16,6 +16,7 @@ import net.dv8tion.jda.core.entities.VoiceChannel;
 import org.cascadebot.cascadebot.data.managers.PlaylistManager;
 import org.cascadebot.cascadebot.data.objects.Playlist;
 import org.cascadebot.cascadebot.data.objects.PlaylistType;
+import org.cascadebot.cascadebot.events.PlayerListener;
 import org.cascadebot.cascadebot.utils.StringsUtil;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
```diff
@@ -73,8 +74,12 @@ public void addTrack(AudioTrack track) {
         }
     }
 
-    public void loop(LoopType loop) {
-        this.loop = loop;
+    public void setLoopMode(LoopMode loopMode) {
```
weeryan17: I prefered this being called loop as all the other methods for controlling the player are named what they do. we could maybe add `loopAll` and `loopSong` methods or just name it back to loop.

binaryoverload: Thing is, this is setting a property, all the other methods are called what they are because they are actions. This is literally `setting the loop mode` and so IMO it should be named as such due to Java convention.

weeryan17: In that case everything else should be named differently witch doesn't make much sense either

binaryoverload: What else should be renamed? I don't see any other methods in this class that go against what I described

weeryan17: Oh right i removed those methods, but still in this case It makes more sense to use the music term as it is a player imo, and this is why i suggest having other methods that make more sense in a music player context like loopAll

binaryoverload: That will then require extra work on the command side, how about just `loopMode` which is a comprise between the two 😛

weeryan17: That works

binaryoverload: Wish we could use kotlin properties xD

PR #86, opened 2019-04-03
Playing command

Ikinon:
# Pull request

 - BinaryOverload :)

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/commands/music/PlayingCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/music/PlayingCommand.java
@@ -8,6 +8,7 @@ import org.cascadebot.cascadebot.commandmeta.CommandContext;
 import org.cascadebot.cascadebot.commandmeta.ICommandMain;
 import org.cascadebot.cascadebot.commandmeta.Module;
 import org.cascadebot.cascadebot.messaging.MessagingObjects;
+import org.cascadebot.cascadebot.music.CascadePlayer;
 import org.cascadebot.cascadebot.permissions.CascadePermission;
 import org.cascadebot.cascadebot.utils.FormatUtils;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/cascadebot/cascadebot/commands/music/PlayingCommand.java
```diff
@@ -0,0 +1,64 @@
...
+    public void onCommand(Member sender, CommandContext context) {
+        IPlayer player = context.getData().getMusicPlayer().getPlayer();
+        AudioTrack track = player.getPlayingTrack();
+
+        if (player.getPlayingTrack() == null) {
+            context.getTypedMessaging().replyWarning("No music playing!");
+        } else {
+            EmbedBuilder embedBuilder = MessagingObjects.getClearThreadLocalEmbedBuilder();
+            embedBuilder.setTitle(track.getInfo().title + " - " + track.getInfo().author, track.getInfo().uri);
+            embedBuilder.setThumbnail("https://img.youtube.com/vi/" + track.getIdentifier() + "/hqdefault.jpg");
```
weeryan17: you don't always have youtube tracks........

src/main/java/org/cascadebot/cascadebot/commands/music/PlayingCommand.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public void onCommand(Member sender, CommandContext context) {
+        IPlayer player = context.getData().getMusicPlayer().getPlayer();
+        AudioTrack track = player.getPlayingTrack();
+
+        if (player.getPlayingTrack() == null) {
+            context.getTypedMessaging().replyWarning("No music playing!");
+        } else {
+            EmbedBuilder embedBuilder = MessagingObjects.getClearThreadLocalEmbedBuilder();
+            embedBuilder.setTitle(track.getInfo().title + " - " + track.getInfo().author, track.getInfo().uri);
```
weeryan17: We can wet embed author why don't we set that to the author of the song

src/main/java/org/cascadebot/cascadebot/commands/music/PlayingCommand.java
```diff
@@ -0,0 +1,64 @@
...
+            embedBuilder.setThumbnail("https://img.youtube.com/vi/" + track.getIdentifier() + "/hqdefault.jpg");
+            embedBuilder.addField("Status", player.isPaused() ? "\u23F8 Paused" : "\u25B6 Playing", true);
+            embedBuilder.addField("Duration", track.getInfo().isStream ? "This is a livestream!" : FormatUtils.formatLongTimeMills(track.getDuration()), true);
+            embedBuilder.addField("Amount played", FormatUtils.formatLongTimeMills(track.getPosition()), true);
```
weeryan17: You should use the progress bar method i have instead of these two, and then add another field thats amount played and make it be `amount/duration` also you got the amount payed wrong you need to get the amount played dirrectly from the player not from the track.

weeryan17:
BTW please start actually using the pr template. it's their for a reason, if you don't use it in the future we **WILL** close your pr.

binaryoverload:
Hey Ryan you see that I signed off on it? The PR template has nothing to do with him

src/main/java/org/cascadebot/cascadebot/commands/music/PlayingCommand.java
binaryoverload: Oof true

src/main/java/org/cascadebot/cascadebot/commands/music/PlayingCommand.java
binaryoverload: The get position works fine according to testing?

weeryan17: With lavaplayer yes as, lavalink no. I ran into this on flare.

src/main/java/org/cascadebot/cascadebot/commands/music/PlayingCommand.java
binaryoverload: I suggest we make a util method somewhere that uses the uri to get what service it is from and therefore what image url to use. Perhaps `MusicUtils`?

weeryan17: I'll look into what services we can get an image from, it might not be worth it if youtube is the only one.

binaryoverload: Well if YouTube is the only one then we can just do it for that

weeryan17: Feels like a waste of a util method.

binaryoverload: Well we could probably just make it into logic in the command if it's just youtube although I doubt it will be just YouTube

PR #88, opened 2019-04-05
Volume Command

Ikinon:
# Pull request

## Feature
Music - Volume Setting

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
@@ -1,0 +1,55 @@
+package org.cascadebot.cascadebot.commands.music;
+
+import net.dv8tion.jda.core.entities.Member;
+import org.cascadebot.cascadebot.commandmeta.CommandContext;
+import org.cascadebot.cascadebot.commandmeta.ICommandMain;
+import org.cascadebot.cascadebot.commandmeta.Module;
+import org.cascadebot.cascadebot.music.CascadePlayer;
+import org.cascadebot.cascadebot.permissions.CascadePermission;
+
+public class VolumeCommand implements ICommandMain {
+
+    @Override
+    public void onCommand(Member sender, CommandContext context) {
+        CascadePlayer player = context.getData().getMusicPlayer();
+        if (context.getArgs().length == 0) {
+            context.getTypedMessaging().replyInfo("Current volume is " + player.getPlayer().getVolume());
+            return;
+        }
+        Integer Arguments = Integer.valueOf(context.getMessage(0));
+        if (Arguments > 100 || Arguments < 0) {
+            context.getTypedMessaging().replyWarning("Volume needs to be between 100 and 0");
+            return;
+        }
+
+        if (Arguments == context.getData().getMusicPlayer().getPlayer().getVolume()) {
+            context.getTypedMessaging().replyInfo("Volume is already " + player.getPlayer().getVolume());
+        } else {
+            player.getPlayer().setVolume(Arguments);
+            context.getTypedMessaging().replyInfo("Volume set to " + player.getPlayer().getVolume());
+        }
+
+    }
+
+
+    @Override
+    public Module getModule() {
+        return Module.MUSIC;
+    }
+
+    @Override
+    public String command() {
+        return "volume";
+    }
+
+    @Override
+    public CascadePermission getPermission() {
+        return CascadePermission.of("Volume", "search", getModule());
+    }
+
+    @Override
+    public String description() {
+        return "Changes the players volume";
+    }
+
+}
--- a/src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
@@ -17,7 +17,7 @@ public class VolumeCommand implements ICommandMain {
             return;
         }
         Integer Arguments = Integer.valueOf(context.getMessage(0));
-        if (Arguments > 100 || Arguments < 0) {
+        if (Arguments < 0 || Arguments > 100) {
             context.getTypedMessaging().replyWarning("Volume needs to be between 100 and 0");
             return;
         }
--- a/src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
@@ -1,11 +1,15 @@
 package org.cascadebot.cascadebot.commands.music;
 
 import net.dv8tion.jda.core.entities.Member;
+import org.cascadebot.cascadebot.commandmeta.Argument;
+import org.cascadebot.cascadebot.commandmeta.ArgumentType;
 import org.cascadebot.cascadebot.commandmeta.CommandContext;
 import org.cascadebot.cascadebot.commandmeta.ICommandMain;
 import org.cascadebot.cascadebot.commandmeta.Module;
 import org.cascadebot.cascadebot.music.CascadePlayer;
 import org.cascadebot.cascadebot.permissions.CascadePermission;
+
+import java.util.Set;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public void onCommand(Member sender, CommandContext context) {
+        CascadePlayer player = context.getData().getMusicPlayer();
+        if (context.getArgs().length == 0) {
+            context.getTypedMessaging().replyInfo("Current volume is %s%%", player.getPlayer().getVolume());
+            return;
+        }
+        Integer Arguments = Integer.valueOf(context.getMessage(0));
+        if (Arguments > 100 || Arguments < 0) {
+            context.getTypedMessaging().replyWarning("Volume needs to be between 100 and 0");
```
weeryan17: flip this around. aka make it `0 and 100` not `100 and 0`

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
```diff
@@ -0,0 +1,55 @@
...
+public class VolumeCommand implements ICommandMain {
+
+    @Override
+    public void onCommand(Member sender, CommandContext context) {
+        CascadePlayer player = context.getData().getMusicPlayer();
+        if (context.getArgs().length == 0) {
+            context.getTypedMessaging().replyInfo("Current volume is %s%%", player.getPlayer().getVolume());
+            return;
+        }
+        Integer Arguments = Integer.valueOf(context.getMessage(0));
```
weeryan17: 1. why is this name Argument`s`
2. camelCase

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public String command() {
+        return "volume";
+    }
+
+    @Override
+    public CascadePermission getPermission() {
+        return CascadePermission.of("Volume", "search", getModule());
```
weeryan17: Thats not how this works. the second argument is the perm so switch that to volume as well

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
weeryan17: oh also you should parse this better.
first off make the variable an int second off parse it with Integer.parseInt() and surround that in a try catch

binaryoverload:
You need to use the method that specifys the flag to restrict this to Patreon guilds

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
Ikinon: `why is this name Arguments` I don't know what else to name it to. Suggestion?

binaryoverload: Why is getMessage being used? We have a get arg as int method

weeryan17: also you can name it `volume` :P

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
```diff
@@ -0,0 +1,62 @@
...
+public class VolumeCommand implements ICommandMain {
+
+    @Override
+    public void onCommand(Member sender, CommandContext context) {
+        CascadePlayer player = context.getData().getMusicPlayer();
+        if (context.getArgs().length == 0) {
+            context.getTypedMessaging().replyInfo("Current volume is %s%%", player.getPlayer().getVolume());
+            return;
+        }
+        Integer Volume = null;
```
weeryan17: 1. int not Integer
2. camelCase

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
```diff
@@ -16,23 +16,21 @@ public void onCommand(Member sender, CommandContext context) {
...
-
-        } catch (java.lang.NumberFormatException e) {
+        int volume = -1;
+        if(context.isArgInteger(0)) {
+            volume = context.getArgAsInteger(0);
             context.getUIMessaging().replyUsage(this);
             return;
         }
-        if (Volume < 0 || Volume > 100) {
+        if (volume < 0 || volume > 100) {
```
binaryoverload: Why don't we just let them go 150 or smth?

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
```diff
@@ -16,23 +16,21 @@ public void onCommand(Member sender, CommandContext context) {
...
             context.getTypedMessaging().replyWarning("Volume needs to be between 100 and 0");
```
binaryoverload: Swap these two round i.e `between 0 and 100`

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
```diff
@@ -16,23 +16,21 @@ public void onCommand(Member sender, CommandContext context) {
...
             return;
         }
 
-        if (Volume == context.getData().getMusicPlayer().getPlayer().getVolume()) {
+        if (volume == context.getData().getMusicPlayer().getPlayer().getVolume()) {
             context.getTypedMessaging().replyInfo("Volume is already %s%%", player.getPlayer().getVolume());
         } else {
-            player.getPlayer().setVolume(Volume);
+            player.getPlayer().setVolume(volume);
             context.getTypedMessaging().replyInfo("Volume set to %s%%", player.getPlayer().getVolume());
```
binaryoverload: Use `%d` this is a number not a string

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
```diff
@@ -16,23 +16,21 @@ public void onCommand(Member sender, CommandContext context) {
...
         }
-        if (Volume < 0 || Volume > 100) {
+        if (volume < 0 || volume > 100) {
             context.getTypedMessaging().replyWarning("Volume needs to be between 100 and 0");
             return;
         }
 
-        if (Volume == context.getData().getMusicPlayer().getPlayer().getVolume()) {
+        if (volume == context.getData().getMusicPlayer().getPlayer().getVolume()) {
             context.getTypedMessaging().replyInfo("Volume is already %s%%", player.getPlayer().getVolume());
```
binaryoverload: Same as below

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
Ikinon: Umm, is it not that already?

weeryan17: in the text not the if statment

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
weeryan17: I think we can't. let me double check though.

src/main/java/org/cascadebot/cascadebot/commands/music/VolumeCommand.java
weeryan17: In fact switch the if statement back around because now thats broken

binaryoverload: The if statement seems fine to me?

weeryan17: Whops nvm

PR #91, opened 2019-04-07
Permissions!

binaryoverload:
# Pull request

## Added/Changed
 - Permissions checking system
   - Most restrictive and hierarchical modes
   - Security levels

## Todo
 - [x] Discord role linking
 - [x] Check discord perms on permissions


src/main/java/org/cascadebot/cascadebot/data/objects/GuildPermissions.java
```diff
@@ -5,9 +5,110 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
 public class GuildPermissions {
 
+    private PermissionMode mode = PermissionMode.HIERARCHICAL;
```
weeryan17: Isn't the default going to be `MOST_RESTRICTIVE` as it's easier to work with?

src/main/java/org/cascadebot/cascadebot/data/objects/GuildPermissions.java
```diff
@@ -5,9 +5,110 @@
...
+    }
 
+    public Group createGroup(String name) {
+        Set<String> ids = groups.stream().map(Group::getId).collect(Collectors.toSet());
+        Group group;
+        int iterations = 0;
+        do {
+            group = new Group(name);
+            if (++iterations == 7) {
+                // If this happens then... run?
```
weeryan17: We should maybe log this lol

src/main/java/org/cascadebot/cascadebot/data/objects/GuildPermissions.java
binaryoverload: Could be, I'd like a discussion about this

weeryan17:
I'm wating for the next metting to review so we can talk about some stuff

src/main/java/org/cascadebot/cascadebot/data/objects/GuildPermissions.java
```diff
@@ -5,9 +5,179 @@
...
+
+    public Group createGroup(String name) {
+        Set<String> ids = groups.stream().map(Group::getId).collect(Collectors.toSet());
+        Group group;
+        int iterations = 0;
+        do {
+            group = new Group(name);
+            if (++iterations == 7) {
+                // If this happens then... run?
+                CascadeBot.LOGGER.error("Somehow we couldn't manage to create a unique guild ID :(");
```
weeryan17: "Somehow we couldn't manage to create a unique **guild** ID :("

might want to fix that lol

src/main/java/org/cascadebot/cascadebot/permissions/objects/PermissionMode.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2019 CascadeBot. All rights reserved.
+ * Licensed under the MIT license.
+ */
+
+package org.cascadebot.cascadebot.permissions.objects;
+
+public enum PermissionMode {
```
weeryan17: Should we make this a sub class of GuildPermission as theirs not a hole lot here?

src/main/java/org/cascadebot/cascadebot/permissions/objects/ResultCause.java
```diff
@@ -0,0 +1,31 @@
+package org.cascadebot.cascadebot.permissions.objects;
+
+public enum ResultCause {
+
+    /**
+     * Indicates that this result originated from a permission in a group.
+     */
+    GROUP,
```
weeryan17: we should maybe re-name these to something like DENY_GROUP in case we end up having a deny and allow needing the same name, and also organizes it better.

src/main/java/org/cascadebot/cascadebot/permissions/objects/Result.java
```diff
@@ -0,0 +1,88 @@
+package org.cascadebot.cascadebot.permissions.objects;
+
+import org.cascadebot.cascadebot.permissions.CascadePermission;
+
+public class Result {
```
weeryan17: Do we really need this class? As far as i can tell it doesn't provide any info that we don't already have access to when checking permissions in the first place, besides for ResultCause and we can just return that.

src/main/java/org/cascadebot/cascadebot/commandmeta/CommandContext.java
```diff
@@ -297,8 +297,13 @@ public String globalEmote(String key) {
     }
 
     public boolean hasPermission(String permission) {
+        return hasPermission(member, permission);
+    }
+
+    public boolean hasPermission(Member member, String permission) {
         CascadePermission cascadePermission = CascadeBot.INS.getPermissionsManager().getPermission(permission);
-        return cascadePermission != null; // TODO: Check actual perms
+        if (cascadePermission == null) return false;
```
weeryan17: we should maybe log this or throw an exception as this should only happen when they send in the wrong permission string

PR #95, opened 2019-04-22
Tags! 🎉

### LINKED ISSUE #93: Tags command
And so we introduce the lovely tags system or "text-based custom commands" as @weeryan17 likes to call it.

This is effectively a system that maps a key to text. A common use for this is rules, knowledge base and support messages for servers along with much more!

Alongside this, we will be adding a category system that will allow tags to be split into different categories such as moderation, knowledge base etc. These categories will not be used anywhere else and will be checked and created dynamically with tags.
```java
class Tag {
    String content
    String category
}
```

There will be a `Map<String, Tag>` added to `GuildData` that will store all the tags for a guild. The appropriate methods should be added to add/remove tags and get the map of tags (immutable!)

A setting should be added to `GuildSettings` called `allowTagCommands` which enables or disables doing `;<tagname>`; this should be enabled by default. The tags commands will be processed after the built in commands in order to avoid conflicts.

A tag command should be created with different usages:
 - `;tag create <tagname> <content>`
 - `;tag create <tagname[:category]> <content>`
 - `;tag delete <tagname>`
 - `;tag <tagname`
 - `;tag list`

JamesHawkinss:
**WIP** This will be a gradual update of my progress on the Tag feature.

# Pull request

## Added/Changed
Added the map to guildData https://thedudeinthecorner.xyz/files/P1GVM.png 🗡

Closes #93


src/main/java/org/cascadebot/cascadebot/commandmeta/Module.java
```diff
@@ -33,6 +33,10 @@
     /**
      * All our special commands :D
      */
+    TAGS,
+    /**
+     * This is a temporary module for housing tags whilst in development.
+     */
```
weeryan17: Ummmm that should go above not bellow the tags module

src/main/java/org/cascadebot/cascadebot/commands/subcommands/tag/TagCreateSubCommand.java
```diff
@@ -0,0 +1,38 @@
...
+import net.dv8tion.jda.core.entities.Member;
+import org.cascadebot.cascadebot.commandmeta.CommandContext;
+import org.cascadebot.cascadebot.commandmeta.ICommandExecutable;
+import org.cascadebot.cascadebot.permissions.CascadePermission;
+
+public class TagCreateSubCommand implements ICommandExecutable {
+
+    @Override
+    public void onCommand(Member sender, CommandContext context) {
+        context.getData().addTag(context.getArg(0), tag);
```
weeryan17: The tag variable doesn't exists.

src/main/java/org/cascadebot/cascadebot/commands/subcommands/tag/TagDeleteSubCommand.java
```diff
@@ -0,0 +1,38 @@
...
+import net.dv8tion.jda.core.entities.Member;
+import org.cascadebot.cascadebot.commandmeta.CommandContext;
+import org.cascadebot.cascadebot.commandmeta.ICommandExecutable;
+import org.cascadebot.cascadebot.permissions.CascadePermission;
+
+public class TagDeleteSubCommand implements ICommandExecutable {
+
+    @Override
+    public void onCommand(Member sender, CommandContext context) {
+        context.getData().removeTag(context.getMessage(0));
```
weeryan17: This should be getArg as we're using getArg above

src/main/java/org/cascadebot/cascadebot/data/objects/Tag.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class Tag {
+
+    private static final Map<String, BiFunction<CommandContext, String[], String>> placeholders = new HashMap<>();
```
weeryan17: I feel like these should be stored separately.

src/main/java/org/cascadebot/cascadebot/data/objects/Tag.java
```diff
@@ -0,0 +1,99 @@
...
+        addPlaceholder("server_region", (context -> context.getGuild().getRegion().getName()));
+        addPlaceholder("server_name", (context -> context.getGuild().getName()));
+        addPlaceholder("server_owner", (context -> context.getGuild().getOwner().getEffectiveName()));
+        addPlaceholder("member_count", (context -> String.valueOf(context.getGuild().getMembers().size())));
+
+        // Sender
+        addPlaceholder("sender", (context -> context.getMember().getEffectiveName()));
+        addPlaceholder("sender_id", (context -> context.getUser().getId()));
+        addPlaceholder("sender_mention", (context -> context.getMember().getAsMention()));
+
```
weeryan17: we should have one of these for getting a different user.

src/main/java/org/cascadebot/cascadebot/data/objects/Tag.java
binaryoverload: If you have any suggestions I'm all ears :)

src/main/java/org/cascadebot/cascadebot/commandmeta/Module.java
binaryoverload: ^ this
also I don't think we should have a tags module as the only thing that's gonna be in it is the tags command. Pretty pointless imo

src/main/java/org/cascadebot/cascadebot/data/objects/Tag.java
weeryan17: Maybe a TagsPlaceholders class as having it in tags will crowd it.

src/main/java/org/cascadebot/cascadebot/data/objects/Tag.java
binaryoverload: How would that work, by id or name? I'm interested to see your use case for that

src/main/java/org/cascadebot/cascadebot/data/objects/Tag.java
binaryoverload: Could do, it would make sense if we're gonna do that to do it in an enum perhaps

src/main/java/org/cascadebot/cascadebot/data/objects/Tag.java
weeryan17: We could hook into the utils we allready have for it. although thinking about it this might be better suited for the full custom command system so the user can better handle it when a user can't be found.

src/main/java/org/cascadebot/cascadebot/data/objects/Tag.java
weeryan17: Probably

src/main/java/org/cascadebot/cascadebot/data/objects/Tag.java
binaryoverload: Yeah, I don't really see why you'd necessarily want to tag another user. We could add a {user:id} tag that would get a user by static id

weeryan17: The main reason is having a tag with info or like a solution to a common issue on like a support server and you want to ping the user having issues. although like i said this would be more suited to the full blown custom commands.

binaryoverload:
jenkins build

binaryoverload:
jenkins build

binaryoverload:
@weeryan17 James has expressed that he doesn't feel he can finish this PR. Do you want to take it over?

weeryan17:
> 
> 
> @weeryan17 James has expressed that he doesn't feel he can finish this PR. Do you want to take it over?

sure

weeryan17:
Closing this pr as I'm taking it over with pr #103

PR #96, opened 2019-04-23
improve saving and loading of playlists

weeryan17:
# Pull request

## Added/Changed
I added return types and consumers to saving and loading playlists so we can have better detection in commands and give users more options.

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
+++ b/src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
@@ -210,7 +210,7 @@ public class CascadePlayer {
         Playlist guild = PlaylistManager.getPlaylistByName(sender.getGuild().getIdLong(), PlaylistType.GUILD, name);
         Playlist user = PlaylistManager.getPlaylistByName(sender.getUser().getIdLong(), PlaylistType.USER, name);
         if(guild != null && user != null) {
-            consumer.accept(LoadPlaylistResult.EXITS_IN_ALL_SCOPES, null);
+            consumer.accept(LoadPlaylistResult.EXISTS_IN_ALL_SCOPES, null);
         } else if (guild != null) {
             loadLoadedPlaylist(guild, sender.getUser().getIdLong(), tracks -> {
                 consumer.accept(LoadPlaylistResult.LOADED_GUILD, tracks);
--- a/src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
+++ b/src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
@@ -257,12 +257,13 @@ public class CascadePlayer {
                 playlist.removeTrack(url);
             }, exception -> {
                 playlist.removeTrack(url);
-            }, tracks::addAll);
+            }, loadedTracks -> {
+                tracks.addAll(loadedTracks);
+                if (tracks.size() == playlist.getTracks().size()) {
+                    loadedConsumer.accept(tracks);
+                }
+            });
         }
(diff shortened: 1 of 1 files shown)


src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
```diff
@@ -228,17 +272,41 @@ public void saveCurrentPlaylist(long owner, PlaylistType scope, String name) {
 
         Playlist search = PlaylistManager.getPlaylistByName(owner, scope, name);
         if (search != null) {
-            search.setTracks(ids);
-            PlaylistManager.replacePlaylist(search);
+            if(overwrite) {
```
binaryoverload: Format
```suggestion
            if (overwrite) {
```

src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
```diff
@@ -228,17 +272,41 @@ public void saveCurrentPlaylist(long owner, PlaylistType scope, String name) {
...
+        OVERWRITE,
+        NEW
+
+    }
+
+    public enum LoadPlaylistResult {
+
+        LOADED_GUILD,
+        LOADED_USER,
+        EXITS_IN_ALL_SCOPES,
```
binaryoverload:
```suggestion
        EXISTS_IN_ALL_SCOPES,
```

src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
```diff
@@ -204,19 +206,61 @@ public void loadFailed(FriendlyException e) {
         });
     }
 
-    public void loadPlaylist(Playlist playlist, long reqUser) {
+    public void loadPlaylist(String name, Member sender, BiConsumer<LoadPlaylistResult, List<AudioTrack>> consumer) {
```
binaryoverload: Is there some way we could use a return value instead of a consumer to make it consistent with save?

weeryan17: Nope because loading the playlists requires a consumer.

binaryoverload: What do you mean down the line?

weeryan17: The `loadLoadedPlaylist` method. Actually thinking about it. I need to change that

src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
```diff
@@ -251,13 +251,18 @@ public void loadPlaylist(String name, Member sender, PlaylistType scope, BiConsu
...
+        List<AudioTrack> tracks = new ArrayList<>();
         for (String url : playlist.getTracks()) {
             loadLink(url, reqUser, noMatch -> {
                 playlist.removeTrack(url);
             }, exception -> {
                 playlist.removeTrack(url);
-            }, loadedConsumer);
+            }, tracks::addAll);
+        }
+        while (tracks.size() < playlist.getTracks().size()) {
```
binaryoverload: I really don't like this, find a better way to do it

weeryan17: be my guest

binaryoverload: This also will infinite loop if any of them fail 🤔

weeryan17: We remove them from the playlist if they fail.

PR #97, opened 2019-04-23
Queue Command

Ikinon:
# Pull request

## Feature
Music - Queue Command

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/commands/music/QueueCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/music/QueueCommand.java
@@ -16,8 +16,10 @@ import org.cascadebot.cascadebot.messaging.MessagingObjects;
 import org.cascadebot.cascadebot.music.CascadePlayer;
 import org.cascadebot.cascadebot.permissions.CascadePermission;
 import org.cascadebot.cascadebot.utils.pagination.Page;
+import org.cascadebot.cascadebot.utils.pagination.PageObjects;
 import org.cascadebot.cascadebot.utils.pagination.PageUtils;
 
+import java.util.ArrayList;
 import java.util.List;
 
 public class QueueCommand implements ICommandMain {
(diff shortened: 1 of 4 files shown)


src/main/java/org/cascadebot/cascadebot/commands/music/QueueCommand.java
```diff
@@ -0,0 +1,80 @@
...
+            builder.append(i).append(": `").append(track.getInfo().title).append("` Requested by ")
+                    .append(CascadeBot.INS.getShardManager().getUserById((Long) track.getUserData()).getAsTag()).append("\n");
+        }
+
+        List<Page> pages;
+
+        if (context.getSettings().useEmbedForMessages()) {
+            pages = PageUtils.splitStringToEmbedPages(builder.toString(), 1800, '\n');
+        } else {
+            pages = PageUtils.splitStringToStringPages(builder.toString(), 1800, '\n');
```
binaryoverload: Instead of doing this I say we just have 10 or 15 entries per page

src/main/java/org/cascadebot/cascadebot/commands/music/QueueCommand.java
```diff
@@ -0,0 +1,80 @@
...
+        StringBuilder builder = new StringBuilder();
+
+        builder.append("**Current song**: `").append(player.getPlayer().getPlayingTrack().getInfo().title).append("` Requested by ")
+                .append(CascadeBot.INS.getShardManager().getUserById((Long) player.getPlayer().getPlayingTrack().getUserData()).getAsTag()).append("\n\n");
+
+        int i = 0;
+        for (AudioTrack track : player.getQueue()) {
+            i++;
+            builder.append(i).append(": `").append(track.getInfo().title).append("` Requested by ")
+                    .append(CascadeBot.INS.getShardManager().getUserById((Long) track.getUserData()).getAsTag()).append("\n");
```
binaryoverload: We should append `\n\n` to add a blank line in between each

src/main/java/org/cascadebot/cascadebot/commands/music/QueueCommand.java
weeryan17: I think that's too little as i see two ways of this being used. 1. a quick glance at whats coming up, 2. seeing the entire playlist and maybe finding a specific song, and we could cover 1 but adding a on deck feature to the song command, otherwise people are just going to have to go though several pages to find something and also keeps everything one 1 page for smaller queues.

binaryoverload: Still, I think even at 10/15 items you're still gonna be able to get between pages quite quickly. 

Music management is another feature which could make it onto the website? That would make viewing whole playlists and management easier.

weeryan17: We don't want to require the web panel to do stuff. we've talked about this before, and I know their are going to be people and guilds with over 100 songs in playlists/queue up at once. It would take 10 seconds to look at the last page (because of rate limits) if we limited it to 10.

PR #117, opened 2019-06-11
Rename "CascadeBot" to "Cascade"

binaryoverload:


JamesHawkinss:
Was this discussed in a meeting?

binaryoverload:
Nope, wondering if we should

JamesHawkinss:
Perhaps. It's a big change to make, after all.

Ikinon:
I don't mind, but can we at least make the "cascadebot" change to "cascade" in the wrapper as well if we're doing it here

JamesHawkinss:
Waiting on meeting.

PR #132, opened 2019-06-16
Fix/guildinfo 

Ikinon:
# Pull request

## Added/Changed
Minor changes to guild command to reduce errors

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
@@ -28,7 +28,7 @@ public class ServerInfoCommand implements ICommandMain {
             try {
                 guildForInfo = CascadeBot.INS.getShardManager().getGuildById(context.getArg(0));
             } catch (NumberFormatException e) {
-                context.getUIMessaging().replyUsage();
+                context.getTypedMessaging().replyDanger(context.i18n("responses.invalid_guild_id"));
                 return;
             }
         }
--- a/src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
@@ -50,7 +50,6 @@ public class ServerInfoCommand implements ICommandMain {
         context.getTypedMessaging().replyInfo(builder); // Send the embed
     }
 
-
     @Override
     public String command() {
         return "serverinfo";
--- a/src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
@@ -29,7 +29,12 @@ public class ServerInfoCommand implements ICommandMain {
         }
 
         if (context.getArgs().length > 0) {
-            guildForInfo = CascadeBot.INS.getShardManager().getGuildById(context.getArg(0));
+            try {
+                guildForInfo = CascadeBot.INS.getShardManager().getGuildById(context.getArg(0));
+            } catch (NumberFormatException e) {
+                context.getUIMessaging().replyUsage();
+                return;
+            }
         }
         if (guildForInfo == null) {
             context.getTypedMessaging().replyDanger(context.i18n("responses.cannot_find_guild"));
--- a/src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
@@ -23,6 +23,10 @@ public class ServerInfoCommand implements ICommandMain {
     @Override
     public void onCommand(Member sender, CommandContext context) {
         Guild guildForInfo = context.getGuild();
+        if (context.getArgs().length != 0 && !context.isArgInteger(0)) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
```diff
@@ -23,9 +23,18 @@
     @Override
     public void onCommand(Member sender, CommandContext context) {
         Guild guildForInfo = context.getGuild();
+        if (context.getArgs().length != 0 && !context.isArgInteger(0)) {
```
binaryoverload: This will make ;serverinfo on it's own fail. Since you've added a check below, you can just remove this

Ikinon: guildinfo on it's own does not fail, but you're right its not needed 👍

src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
```diff
@@ -25,7 +25,12 @@ public void onCommand(Member sender, CommandContext context) {
         Guild guildForInfo = context.getGuild();
 
         if (context.getArgs().length > 0) {
-            guildForInfo = CascadeBot.INS.getShardManager().getGuildById(context.getArg(0));
+            try {
+                guildForInfo = CascadeBot.INS.getShardManager().getGuildById(context.getArg(0));
+            } catch (NumberFormatException e) {
```
weeryan17: ummmm what? we're not doing any conversion to numbers here why do you have this

src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
```diff
@@ -40,11 +45,12 @@ public void onCommand(Member sender, CommandContext context) {
         builder.addField(context.i18n("commands.serverinfo.guild_owner") + context.globalEmote("server_owner"), guildForInfo.getOwner().getUser().getAsTag(), true);
         builder.addField(context.i18n("commands.serverinfo.guild_region"), guildForInfo.getRegion().toString(), true);
         builder.addField(context.i18n("commands.serverinfo.guild_member_count"), String.valueOf(guildForInfo.getMemberCache().size()), true);
-        builder.setFooter(context.i18n("commands.serverinfo.guild_id", context.getGuild().getId()) + guildForInfo.getId(), guildForInfo.getIconUrl());
+        builder.setFooter(context.i18n("commands.serverinfo.guild_id", guildForInfo.getId()), guildForInfo.getIconUrl());
 
         context.getTypedMessaging().replyInfo(builder); // Send the embed
     }
 
+
```
weeryan17: random whitespace

src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
binaryoverload: getGuildById does actually throw number format exception if the input is not a long

weeryan17: really? i though it just returned null

binaryoverload: Nope

src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
```diff
@@ -25,7 +25,12 @@ public void onCommand(Member sender, CommandContext context) {
         Guild guildForInfo = context.getGuild();
 
         if (context.getArgs().length > 0) {
-            guildForInfo = CascadeBot.INS.getShardManager().getGuildById(context.getArg(0));
+            try {
+                guildForInfo = CascadeBot.INS.getShardManager().getGuildById(context.getArg(0));
+            } catch (NumberFormatException e) {
+                context.getUIMessaging().replyUsage();
```
weeryan17: Why are you replying usage? you should say they gave an invalid guild id.

src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
binaryoverload: We really need better utils for this but I'm gonna look at that for beta

src/main/java/org/cascadebot/cascadebot/commands/informational/ServerInfoCommand.java
Ikinon: "We cannot find that guild" or are we going to be specific here on that its guild ID

weeryan17: You should say something along the lines of they need to use a guild id, as otherwise people will enter a guilds name and then get confused

Ikinon: "We could not find that guild ID!" enough?

weeryan17: I was thinking more along the lines of "That isn't a valid guild id", or "%s isn't a valid guild id" (obviously replacing %s with that the user entered). This isn't the guild not being able to be found this is them entering the wrong information.

PR #135, opened 2019-06-20
Use the abstract discord URL

binaryoverload:
To make sure we can change it without changing every single link


Ikinon:
We should probably change this in the constants file in the bot too then

PR #141, opened 2019-07-06
Catch error on user not found

Ikinon:
# Pull request

## Added/Changed
Changed ban command to catch errors when the user is not found

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/commands/moderation/BanCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/moderation/BanCommand.java
@@ -8,6 +8,7 @@ package org.cascadebot.cascadebot.commands.moderation;
 import net.dv8tion.jda.core.Permission;
 import net.dv8tion.jda.core.entities.Member;
 import net.dv8tion.jda.core.entities.User;
+import net.dv8tion.jda.core.exceptions.ErrorResponseException;
 import org.cascadebot.cascadebot.CascadeBot;
 import org.cascadebot.cascadebot.commandmeta.Argument;
 import org.cascadebot.cascadebot.commandmeta.ArgumentType;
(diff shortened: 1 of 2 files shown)


src/main/java/org/cascadebot/cascadebot/commands/moderation/BanCommand.java
```diff
@@ -42,14 +43,14 @@ public void onCommand(Member sender, CommandContext context) {
         if (targetMember == null) {
             // If the member is null, the user does not exist in the guild.
             // This attempts to retrieve the user from Discord.
-            targetUser = DiscordUtils.getUser(context.getGuild(), context.getMessage(0), true);
-
-            if (targetUser == null) {
+            try {
+                targetUser = DiscordUtils.getUser(context.getGuild(), context.getMessage(0), true);
+            } catch (ErrorResponseException | NumberFormatException e) {
```
binaryoverload: We should really be editing DiscordUtils to handle this

Ikinon: Yea weeryan said that in the last PR that I did, but you said for beta. Are we doing that here now?

src/main/java/org/cascadebot/cascadebot/utils/DiscordUtils.java
```diff
@@ -67,12 +68,15 @@ public static User getUser(Guild guild, String search, boolean retrieve) {
         Checks.notBlank(search, "user");
         List<User> users = FinderUtil.findUsers(search, guild.getJDA());
         if (users.size() > 1) {
-            /*context.replyDanger("There is more than one user that matches this criterion! Please enter the ID or the user's full name!");*/
```
binaryoverload: When I said uncomment this, I didn't mean delete it 😂

Ikinon: We don't have context passed through so we can't reply like that
*binaryoverload said to comment it back in & say why on discord

PR #145, opened 2019-07-14
Better use of embeds on userinfo

Ikinon:
# Pull request

## Added/Changed
Show requester
Use InfoMessage colour
~~Use embed for roles~~
~~Add pages for when people have a lot of roles~~

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
@@ -7,7 +7,11 @@ package org.cascadebot.cascadebot.commands.informational;
 
 import net.dv8tion.jda.core.EmbedBuilder;
 import net.dv8tion.jda.core.OnlineStatus;
-import net.dv8tion.jda.core.entities.*;
+import net.dv8tion.jda.core.entities.Game;
+import net.dv8tion.jda.core.entities.Member;
+import net.dv8tion.jda.core.entities.RichPresence;
+import net.dv8tion.jda.core.entities.Role;
+import net.dv8tion.jda.core.entities.User;
 import org.cascadebot.cascadebot.CascadeBot;
 import org.cascadebot.cascadebot.Constants;
 import org.cascadebot.cascadebot.commandmeta.CommandContext;
@@ -20,11 +24,10 @@ import org.cascadebot.cascadebot.utils.FormatUtils;
 import org.cascadebot.cascadebot.utils.language.LanguageUtils;
 import org.cascadebot.cascadebot.utils.pagination.Page;
 import org.cascadebot.cascadebot.utils.pagination.PageObjects;
+import org.cascadebot.shared.SecurityLevel;
 
 import java.util.ArrayList;
 import java.util.List;
-
-import org.cascadebot.shared.SecurityLevel;
 
 public class UserInfoCommand implements ICommandMain {
 
@@ -41,7 +44,7 @@ public class UserInfoCommand implements ICommandMain {
         Member member = context.getGuild().getMember(userForInfo);
         String status = "";
         String statusName = "";
-        String footer = "Requested by " + sender.getUser().getAsTag();
+        String footer = MessagingObjects.getMessageTypeEmbedBuilder()
         EmbedBuilder builder = MessagingObjects.getClearThreadLocalEmbedBuilder();
 
         if (member != null) {
--- a/src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
@@ -94,7 +94,7 @@ public class UserInfoCommand implements ICommandMain {
                 builder.addField(context.i18n("commands.userinfo.activity"), gameStatus, true);
             }
 
-            context.reply(builder.build());
+            context.getTypedMessaging().replyInfo(builder);
         }
     }
 
--- a/src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
@@ -44,7 +44,8 @@ public class UserInfoCommand implements ICommandMain {
         Member member = context.getGuild().getMember(userForInfo);
         String status = "";
         String statusName = "";
-        EmbedBuilder builder = new EmbedBuilder(MessagingObjects.getMessageTypeEmbedBuilder(MessageType.INFO, sender.getUser()));
+        EmbedBuilder builder = MessagingObjects.getMessageTypeEmbedBuilder(MessageType.INFO, sender.getUser());
+
 
         if (member != null) {
             builder.addField(context.i18n("commands.userinfo.user_join_date"), FormatUtils.formatDateTime(member.getJoinDate(), context.getLocale()), true);
@@ -63,14 +64,12 @@ public class UserInfoCommand implements ICommandMain {
             }
         }
 
-        List<Page> pageList = new ArrayList<>();
 
         builder.setTitle(userForInfo.getAsTag());
         builder.setThumbnail(userForInfo.getAvatarUrl());
         builder.addField(context.i18n("commands.userinfo.user_created"), FormatUtils.formatDateTime(userForInfo.getCreationTime(), context.getLocale()), true);
         builder.addField(context.i18n("commands.userinfo.user_id"), userForInfo.getId(), true);
         builder.addField(context.i18n("commands.userinfo.user_mutual_servers"), String.valueOf(userForInfo.getMutualGuilds().size()), true);
-        pageList.add(new PageObjects.EmbedPage(builder, false));
         long userId = userForInfo.getIdLong();
         SecurityLevel userSecurityLevel = CascadeBot.INS.getPermissionsManager().getUserSecurityLevel(userId);
         if (userSecurityLevel != null) {
@@ -95,23 +94,8 @@ public class UserInfoCommand implements ICommandMain {
                 builder.addField(context.i18n("commands.userinfo.activity"), gameStatus, true);
             }
 
-            StringBuilder stringer = new StringBuilder();
-            String author = "Roles for: " + sender.getUser().getAsTag();
-            EmbedBuilder embed = new EmbedBuilder(MessagingObjects.getMessageTypeEmbedBuilder(MessageType.INFO, sender.getUser()).setAuthor(author));
-            int i = 1;
-            for (Role role : member.getRoles()) {
-                if (i % 20 == 0) {
-                    pageList.add(new PageObjects.EmbedPage(
-                            new EmbedBuilder(MessagingObjects.getMessageTypeEmbedBuilder(MessageType.INFO, sender.getUser()).setAuthor(author).setAuthor(author)),
-                            false));
-                    stringer = new StringBuilder();
-                }
-                stringer.append("<@&" + role.getId() + ">" + " ID: " + role.getId() + "\n");
-                i = i + 1;
-            }
-            pageList.add(new PageObjects.EmbedPage(embed.setDescription(stringer.toString()), false));
+            context.reply(builder.build());
         }
-        context.getUIMessaging().sendPagedMessage(pageList);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
```diff
@@ -7,29 +7,23 @@
 
 import net.dv8tion.jda.core.EmbedBuilder;
 import net.dv8tion.jda.core.OnlineStatus;
-import net.dv8tion.jda.core.entities.Game;
-import net.dv8tion.jda.core.entities.Member;
-import net.dv8tion.jda.core.entities.RichPresence;
-import net.dv8tion.jda.core.entities.Role;
-import net.dv8tion.jda.core.entities.User;
+import net.dv8tion.jda.core.entities.*;
```
binaryoverload: Nope

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
```diff
@@ -45,9 +39,9 @@ public void onCommand(Member sender, CommandContext context) {
             return;
         }
         Member member = context.getGuild().getMember(userForInfo);
-        
         String status = "";
         String statusName = "";
+        String footer = "Requested by " + sender.getUser().getAsTag();
```
binaryoverload: Remove this and instead get the builder with `MessagingObjects.getMessageTypeEmbedBuilder`

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
```diff
@@ -74,7 +68,9 @@ public void onCommand(Member sender, CommandContext context) {
         builder.addField(context.i18n("commands.userinfo.user_created"), FormatUtils.formatDateTime(userForInfo.getCreationTime()), true);
         builder.addField(context.i18n("commands.userinfo.user_id"), userForInfo.getId(), true);
         builder.addField(context.i18n("commands.userinfo.user_mutual_servers"), String.valueOf(userForInfo.getMutualGuilds().size()), true);
-
+        builder.setFooter(footer, sender.getUser().getEffectiveAvatarUrl());
+        builder.setColor(Constants.COLOR_INFO);
```
binaryoverload: Remove line 71 and 72, see above

binaryoverload:
Could you also edit the title to be a bit more meaningful?

Ikinon:
> Could you also edit the title to be a bit more meaningful?

Does that do it?

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
```diff
@@ -45,10 +42,9 @@ public void onCommand(Member sender, CommandContext context) {
             return;
         }
         Member member = context.getGuild().getMember(userForInfo);
-        
         String status = "";
         String statusName = "";
-        EmbedBuilder builder = MessagingObjects.getClearThreadLocalEmbedBuilder();
+        EmbedBuilder builder = new EmbedBuilder(MessagingObjects.getMessageTypeEmbedBuilder(MessageType.INFO, sender.getUser()));
```
weeryan17: Use the clear thread local embed builder

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
```diff
@@ -74,7 +70,7 @@ public void onCommand(Member sender, CommandContext context) {
         builder.addField(context.i18n("commands.userinfo.user_created"), FormatUtils.formatDateTime(userForInfo.getCreationTime()), true);
         builder.addField(context.i18n("commands.userinfo.user_id"), userForInfo.getId(), true);
         builder.addField(context.i18n("commands.userinfo.user_mutual_servers"), String.valueOf(userForInfo.getMutualGuilds().size()), true);
-
+        pageList.add(new PageObjects.EmbedPage(builder, false));
```
weeryan17: wat

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
```diff
@@ -99,16 +95,22 @@ public void onCommand(Member sender, CommandContext context) {
...
+            int i = 1;
             for (Role role : member.getRoles()) {
-                tableBuilder.addRow(role.getId(), role.getName());
+                if (i % 20 == 0) {
+                    pageList.add(new PageObjects.EmbedPage(
+                            new EmbedBuilder(MessagingObjects.getMessageTypeEmbedBuilder(MessageType.INFO, sender.getUser()).setAuthor(author).setAuthor(author)),
+                            false));
+                    stringer = new StringBuilder();
+                }
+                stringer.append("<@&" + role.getId() + ">" + " ID: " + role.getId() + "\n");
```
weeryan17: no. just no. we don't want to mention roles every time someone checks userinfo

Ikinon: Pings don't work in embeds...?

weeryan17: That could easily change in the future, and pining in this doesn't make seance when a name would do just fine.

binaryoverload: I disagree. Mentions have never and probably will never work in embeds so I think this is a nice way to do it. It means that we have colour with it as well. Humans work on colour 😄

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
binaryoverload: Just use this instead: (Ryan and I have talked)
```suggestion
        EmbedBuilder builder = MessagingObjects.getMessageTypeEmbedBuilder(MessageType.INFO, sender.getUser());
```

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
weeryan17: I still think they could change it, and also well yes it does show the color that's all it does extra, I'd prefer to give a color value then to show the color. also what was wrong with using the table?

binaryoverload: The tables are broken on mobile; this isn't. I also feel like adding colour is a good thing, and we shouldn't take it away because "that's all it does".

weeryan17: Tables looked fine on mobile for me. in fact embeds are broken on mobile for me atm. I'm still under the opinion mentioning is a bad idea.

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
Ikinon: I'm really not sure what you want me to do with this

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
Ikinon: Also to get to that page, we do edit the message which is something else that does not ping
Anyways I'll let you 2 come to a conclusion about this

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
weeryan17: I'm just wondering why it was moved

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
weeryan17: It pining isn't my only point. a name does just fine. doesn't have the @ symbol in front of it. discord could change it where editing, or embeds will ping in the future. plus like i mentioned embeds are broken on android atm so the table even works better on mobile atm.

binaryoverload: > plus like i mentioned embeds are broken on android atm so the table even works better on mobile atm.

Even so, this should not be a reason as it's only broken temporarily. Tables have always been broken on mobile due to the screen width not being wide enough on some models. Because of this, we should really avoid tables when we can.

>  a name does just fine. doesn't have the @ symbol in front of it. 
The @ symbol is a bit distracting yeah but I still think it looks better than a plain name. Simply because it has the colour that allows you the mentally link the role in discord with the role we are displaying. Especially on guilds where people might not know all the role names, a colour helps link it better.

> discord could change it where editing
Discord could do anything, it's not really our responsibility to try and predict it. 

I'll also add that we are on of the only bots that use tables and the reason for that is because they look so terrible if the screen size isn't big enough (Not so much of an issue here but in larger tables it can be) so I still definitely think that we should avoid tables and find better ways to represent data.

Ikinon: I think it probably depends on more than just how it looks on android alone, because embeds look fine & even tables are almost how they should be on my phone, it'll probably depend on other factors like screen size
/shrug

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
Ikinon: Using that just seems to break the way the pages work, the 1st page about the user is gone and is also replaced with roles

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
weeryan17: Tables work completely fine if they contain smaller sets of data, and are a lot more compact displaying wise. With this change user info will store a lot less info per page. We have a table system why are we avoiding using it for things it's good at like this?
I do agree it would be nice to display the color (along with it's hex code) but that doesn't mean we should compromise other stuff just for that.

binaryoverload: Tables are actually less compact using the method in the PR because we don't have any of the headers or anything like that. Also, it's good at it yeah but, as previously mentioned, we have had reports of users on smaller screen sizes not being able to read it properly. I'm not saying we should avoid the table system for things where we can't do anything else but we **can** display it a different way here. 

What do you think we're compromising with this method? We're displaying the exact same information with the colour as an added bonus.

Also, I don't think we should display a hex code here as it's used to see what roles a user has, not what colour a role is. The `_roles` command (Which we do use a table for and I think that's the best way) is for that.

weeryan17: >Tables are actually less compact using the method in the PR because we don't have any of the headers or anything like that.
not in terms of the amount on a page, not to mention this way things don't line up as nicely

> we have had reports of users on smaller screen sizes not being able to read it properly
for this specific command or just commands in general.
on this specific command cause that makes a huge difference

>What do you think we're compromising with this method? We're displaying the exact same information with the colour as an added bonus.
1. looks less organized as the ids arn't even spaced from the role name
2. the @ symbol is ugly and could be confused for being part of the role name

binaryoverload: For this specific command, from Jip.

I'll give you the alignment and confusion, those are valid points.

src/main/java/org/cascadebot/cascadebot/commands/informational/UserInfoCommand.java
```diff
@@ -77,7 +77,7 @@ public void onCommand(Member sender, CommandContext context) {
         }
 
         if (member != null) {
-            builder.addField("Status", status + statusName, true);
+            builder.addField(context.i18n("commands.userinfo.status"), status + statusName, true);
```
binaryoverload: I have a feeling this was in the words section

binaryoverload: Yeah it is `words.status`, can you delete `commands.userinfo.status` from the lang and change it where it's used

PR #165, opened 2019-07-20
I'm a derp

weeryan17:
What was i thinking when i made this

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/utils/pagination/PageUtils.java
+++ b/src/main/java/org/cascadebot/cascadebot/utils/pagination/PageUtils.java
@@ -122,7 +122,9 @@ public class PageUtils {
                 i = 0;
             }
         }
-        pages.add(new PageObjects.TablePage(tableBuilder.build()));
+        if(tableBuilder.getBody().size() > 0) {
+            pages.add(new PageObjects.TablePage(tableBuilder.build()));
+        }
         return pages;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/cascadebot/cascadebot/utils/pagination/PageUtils.java
```diff
@@ -104,18 +104,25 @@
...
+                pages.add(new PageObjects.TablePage(tableBuilder.build()));
+                tableBuilder = new Table.TableBuilder();
+                for (String head : table.getHeadings()) {
+                    tableBuilder.addHeading(head);
+                }
                 i = 0;
             }
         }
-        pages.add(new PageObjects.TablePage(table));
+        pages.add(new PageObjects.TablePage(tableBuilder.build()));
```
mrjvs: What if there is no overflow 🤔

src/main/java/org/cascadebot/cascadebot/utils/pagination/PageUtils.java
```diff
@@ -104,18 +104,27 @@
...
+                pages.add(new PageObjects.TablePage(tableBuilder.build()));
+                tableBuilder = new Table.TableBuilder();
+                for (String head : table.getHeadings()) {
+                    tableBuilder.addHeading(head);
+                }
                 i = 0;
             }
         }
-        pages.add(new PageObjects.TablePage(table));
+        if(tableBuilder.getBody().size() > 0) {
```
cascadediscordbot:
```suggestion
        if (tableBuilder.getBody().size() > 0) {
```

PR #166, opened 2019-07-25
GraphQL! 🎉

### LINKED ISSUE #163: GraphQL API
We need to create the schema for all the API actions. 

All the guild data will be under one parent guild object similar to how it is in code. 

We're also going to use GraphQL mutations also so an interface for that needs to be created.

binaryoverload:
# Pull request

## Added/Changed
 - GraphQL API
   - The endpoint is `/graphql`
   - The schema is generated dynamically using annotations. To get a copy of the schema (If the one in the root dir is not updated) then use [this tool](https://github.com/graphql-cli/graphql-cli)

## Todo
 - [ ] Config for Spark
 - [ ] Settings - Getting and setting

Closes #163


pom.xml
```diff
@@ -27,6 +27,9 @@
                     <source>11</source>
                     <target>11</target>
                     <release>11</release>
+                    <compilerArguments>
```
weeryan17: wat

src/main/java/org/cascadebot/cascadebot/CascadeBot.java
```diff
@@ -91,7 +96,12 @@ public static void main(String[] args) {
         if (System.getenv("SENTRY_DSN") == null) {
             LOGGER.warn("You haven't set a Sentry DNS in the environment variables! Set SENTRY_DSN to your DSN for this to work!");
         }
-        INS.init();
+        try {
+            INS.init();
+        } catch (Exception e) {
+            LOGGER.error("Error on initialisation!", e);
+            ShutdownHandler.exitWithError();
```
weeryan17: is this the one that tells the wrapper not to restart the bot?

src/main/java/org/cascadebot/cascadebot/permissions/objects/User.java
```diff
@@ -6,6 +6,7 @@
 package org.cascadebot.cascadebot.permissions.objects;
 
 import com.google.common.collect.Sets;
+import io.leangen.graphql.annotations.types.GraphQLType;
```
weeryan17: useless import?

pom.xml
binaryoverload: In order for GraphQL to recognise method parameter names, we need to add this

src/main/java/org/cascadebot/cascadebot/CascadeBot.java
binaryoverload: Yup

pom.xml
weeryan17: again wat. I understand why this is here now but seriously graphql

binaryoverload: It's because normally method parameters don't have reflection metadata so it turns out as `arg0`, `arg1` etc. We could do it without method parameters by using more annotations but I think this is easier and keeps the code cleaner. So it's Java's fault in this case

src/main/java/org/cascadebot/cascadebot/data/managers/PlaylistManager.java
```diff
@@ -41,17 +50,27 @@ public static void savePlaylist(Playlist playlist) {
...
+            );
+        });
+    }
+
+    public static void deletePlaylistById(String id) {
+        CascadeBot.INS.getDatabaseManager().runAsyncTask(database -> {
+            ObjectId playlistId = new ObjectId(id);
+            database.getCollection(COLLECTION, Playlist.class).deleteOne(
+                    eq("_id", playlistId),
+                    new DebugLogCallback<>("Deleted playlist with ID: " + playlistId)
```
mrjvs: 1. change everything to lowercase Id
2. add a log with uppercase ID
3. profit...

binaryoverload: No

weeryan17: /\

binaryoverload:
jenkins build

PR #178, opened 2019-10-01
Better utils for getting numbers from args

### LINKED ISSUE #158: Better utils for getting numbers from args and checking that args are a number.
Using Double/Long/Integer/Float.valueOf() and a try catch seems to be a better option than regex

binaryoverload:
# Pull request

## Added/Changed
 - Made all args checking methods use the `.parseX` methods with exceptions to make sure range is respected
 - The util methods for both checking and parsing are as follows: 
   - Int
   - Long
   - Float
   - Double

Closes #158

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/commandmeta/CommandContext.java
+++ b/src/main/java/org/cascadebot/cascadebot/commandmeta/CommandContext.java
@@ -129,11 +129,41 @@ public class CommandContext {
     }
 
     public boolean isArgInteger(int index) {
-        return Regex.INTEGER_REGEX.matcher(this.args[index]).matches();
+        try {
+            Integer.parseInt(args[index]);
+            return true;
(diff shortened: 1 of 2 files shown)


src/main/java/org/cascadebot/cascadebot/commandmeta/CommandContext.java
```diff
@@ -129,11 +129,41 @@ public String getMessage(int start, int end) {
     }
 
     public boolean isArgInteger(int index) {
-        return Regex.INTEGER_REGEX.matcher(this.args[index]).matches();
```
weeryan17: Remember to remove the regex variable

binaryoverload: That's in Shared, I'll do that now

PR #188, opened 2019-10-23
Feature/persistant buttons

weeryan17:
# Pull request

## Added/Changed
Added extension to the button group for persistent buttons.
stored persistent buttons in guild data.
Moved todo and votes over to persistent.
Made messaging automatically store buttons in the correct places

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/UnicodeConstants.java
+++ b/src/main/java/org/cascadebot/cascadebot/UnicodeConstants.java
@@ -48,6 +48,33 @@ public class UnicodeConstants {
     public static final String EIGHT = "\u0038\u20E3"; // 8️⃣
     public static final String NINE = "\u0039\u20E3"; // 9️⃣
 
+    public static final String A = "\uD83C\uDDE6";
+    public static final String B = "\uD83C\uDDE7";
+    public static final String C = "\uD83C\uDDE8";
+    public static final String D = "\uD83C\uDDE9";
+    public static final String E = "\uD83C\uDDEA";
+    public static final String F = "\uD83C\uDDEB";
+    public static final String G = "\uD83C\uDDEC";
+    public static final String H = "\uD83C\uDDED";
+    public static final String I = "\uD83C\uDDEE";
+    public static final String J = "\uD83C\uDDEF";
+    public static final String K = "\uD83C\uDDF0";
+    public static final String L = "\uD83C\uDDF1";
+    public static final String M = "\uD83C\uDDF2";
+    public static final String N = "\uD83C\uDDF3";
+    public static final String O = "\uD83C\uDDF4";
+    public static final String P = "\uD83C\uDDF5";
+    public static final String Q = "\uD83C\uDDF6";
+    public static final String R = "\uD83C\uDDF7";
+    public static final String S = "\uD83C\uDDF8";
+    public static final String T = "\uD83C\uDDF9";
+    public static final String U = "\uD83C\uDDFA";
+    public static final String V = "\uD83C\uDDFB";
+    public static final String W = "\uD83C\uDDFC";
+    public static final String X = "\uD83C\uDDFD";
+    public static final String Y = "\uD83C\uDDFE";
+    public static final String Z = "\uD83C\uDDFF";
+
     public static final String WHITE_CIRCLE = "\u26AA"; //⚪
     public static final String WHITE_SQUARE = "\uD83D\uDD33"; //⬜
     public static final String WHITE_HALLOW_SQUARE = "\u2B1C"; //🔳
--- a/src/main/java/org/cascadebot/cascadebot/data/database/DatabaseManager.java
+++ b/src/main/java/org/cascadebot/cascadebot/data/database/DatabaseManager.java
@@ -37,7 +37,8 @@ public class DatabaseManager {
             CodecRegistries.fromProviders(PojoCodecProvider.builder().register(
                     "org.cascadebot.cascadebot.data.objects",
                     "org.cascadebot.cascadebot.permissions.objects",
-                    "org.cascadebot.shared"
+                    "org.cascadebot.shared",
+                    "org.cascadebot.cascadebot.utils.buttons"
             ).build())
     );
 
--- a/src/main/java/org/cascadebot/cascadebot/data/database/DatabaseManager.java
+++ b/src/main/java/org/cascadebot/cascadebot/data/database/DatabaseManager.java
@@ -37,8 +37,8 @@ public class DatabaseManager {
             CodecRegistries.fromProviders(PojoCodecProvider.builder().register(
                     "org.cascadebot.cascadebot.data.objects",
                     "org.cascadebot.cascadebot.permissions.objects",
-                    "org.cascadebot.shared",
-                    "org.cascadebot.cascadebot.utils.buttons"
+                    "org.cascadebot.cascadebot.utils.buttons",
+                    "org.cascadebot.shared"
             ).build())
     );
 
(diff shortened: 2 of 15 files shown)


src/main/java/org/cascadebot/cascadebot/UnicodeConstants.java
```diff
@@ -48,6 +48,33 @@
...
+    public static final String Q = "\uD83C\uDDF6";
+    public static final String R = "\uD83C\uDDF7";
+    public static final String S = "\uD83C\uDDF8";
+    public static final String T = "\uD83C\uDDF9";
+    public static final String U = "\uD83C\uDDFA";
+    public static final String V = "\uD83C\uDDFB";
+    public static final String W = "\uD83C\uDDFC";
+    public static final String X = "\uD83C\uDDFD";
+    public static final String Y = "\uD83C\uDDFE";
+    public static final String Z = "\uD83C\uDDFF";
```
binaryoverload: Following the standard in UnicodeConstants, we should probably have a comment with the emoji

src/main/java/org/cascadebot/cascadebot/events/ButtonEventListener.java
```diff
@@ -33,15 +36,25 @@ public void onMessageReactionAdd(MessageReactionAddEvent e) {
...
+                if(data.getPersistentButtons().get(channel.getIdLong()).containsKey(e.getMessageIdLong())) {
+                    ButtonGroup group = data.getPersistentButtons().get(channel.getIdLong()).get(e.getMessageIdLong());
+                    doButtonPress(group, e.getReaction(), e.getMember(), e.getTextChannel());
                 }
             }
         }
     }
 
+    private void doButtonPress(ButtonGroup group, MessageReaction reaction, Member sender, TextChannel channel) {
+        Metrics.INS.buttonsPressed.labels(reaction.getReactionEmote().getName()).inc();
```
binaryoverload: Maybe we could add a label about what type of button this is? i.e. persistent or not

src/main/java/org/cascadebot/cascadebot/utils/buttons/ButtonGroup.java
```diff
@@ -14,14 +15,17 @@
 import net.dv8tion.jda.api.entities.TextChannel;
 import org.cascadebot.cascadebot.CascadeBot;
 
+import javax.naming.OperationNotSupportedException;
```
binaryoverload: Random import?

src/main/java/org/cascadebot/cascadebot/utils/buttons/ButtonGroup.java
```diff
@@ -90,7 +95,7 @@ public void setMessage(long id) {
     }
 
     public void handleButton(Member clicker, TextChannel channel, Message buttonMessage, MessageReaction.ReactionEmote emote) {
-        for (Button button : buttons) {
+        for (Button button : this.getButtons()) {
```
binaryoverload: Why the change? Doesn't make a difference just wondering why

src/main/java/org/cascadebot/cascadebot/utils/buttons/PersistentButtonGroup.java
```diff
@@ -0,0 +1,44 @@
...
+    public List<Button> getButtons() {
+        return persistentButtons.stream().map(PersistentButton::getButton).collect(Collectors.toList());
+    }
+
+    @Override
+    public void addButton(Button button) {
+    }
+
+    @Override
+    public void removeButton(Button button) {
```
binaryoverload: If these are not supposed to be used, they should probably throw an exception inside to make debugging easier. Something like `UnsupportedOperationException` would work perfect in this case

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteMessageType.java
```diff
@@ -20,11 +20,6 @@
     /**
      * This vote type will use discord regional indicator letters up to the amount specified in the builder.
      */
-    LETTERS,
-
-    /**
-     * This vote type will give you complete control over what reactions you want to use .
-     */
-    CUSTOM
+    LETTERS
```
binaryoverload: Where's the custom type gone?

src/main/java/org/cascadebot/cascadebot/utils/buttons/ButtonGroup.java
weeryan17: It actually does make a difference. This allows it to get the correct list of buttons for say persistent buttons. Persistent buttons actually doesn't use the standard button list at all.

src/main/java/org/cascadebot/cascadebot/utils/buttons/PersistentButtonGroup.java
weeryan17: I tried but it wanted me to add a throw clause to the method and therefor the parent method which broke everything.

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteMessageType.java
weeryan17: Because the custom type is hard to do now that they're persistent the way it was being done before. At some point I might return to this.

src/main/java/org/cascadebot/cascadebot/utils/buttons/PersistentButtonGroup.java
binaryoverload: If you use `java.lang.UnsupportedOperationException` it won't force you to try catch since it's a runtime exception. I think before you tried using the javax one which is probably a checked exception

weeryan17: I'll try that

src/main/java/org/cascadebot/cascadebot/utils/buttons/ButtonGroup.java
binaryoverload: Oh I see, this is used so it gets the button list from the overrided method instead. Got it 👌

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroupBuilder.java
```diff
@@ -57,12 +58,12 @@ public VoteButtonGroupBuilder(VoteMessageType type) {
...
     public VoteButtonGroupBuilder setOptionsAmount(int amount) {
-        if (type == VoteMessageType.YES_NO || type == VoteMessageType.CUSTOM) {
-            throw new UnsupportedOperationException("Cannot set options amount for yes no votes, or custom votes");
+        if (type == VoteMessageType.YES_NO) {
+            throw new UnsupportedOperationException("Cannot set options amount for yes no votes");
         }
 
-        if (type == VoteMessageType.LETTERS && amount > 26) {
-            throw new UnsupportedOperationException("Cannot have more then 26 options when using letters");
+        if (type == VoteMessageType.LETTERS && amount > 15) {
```
binaryoverload: Could you just leave a note about why this is 15? Just for future reference (We will forget)

src/main/java/org/cascadebot/cascadebot/data/database/DatabaseManager.java
```diff
@@ -37,7 +37,8 @@
             CodecRegistries.fromProviders(PojoCodecProvider.builder().register(
                     "org.cascadebot.cascadebot.data.objects",
                     "org.cascadebot.cascadebot.permissions.objects",
-                    "org.cascadebot.shared"
+                    "org.cascadebot.shared",
+                    "org.cascadebot.cascadebot.utils.buttons"
```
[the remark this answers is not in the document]
weeryan17: 🤦

src/main/java/org/cascadebot/cascadebot/data/objects/GuildData.java
```diff
@@ -22,11 +22,14 @@
...
 import org.cascadebot.cascadebot.utils.buttons.ButtonGroup;
 import org.cascadebot.cascadebot.utils.buttons.ButtonsCache;
+import org.cascadebot.cascadebot.utils.buttons.PersistentButtonGroup;
 import org.cascadebot.cascadebot.utils.pagination.PageCache;
 
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
```
binaryoverload: Unused import :P

Issue #189, opened 2019-10-25
Uxui Improvements


JamesHawkinss:
# James
- [x] Add aliases to usage command - @JamesHawkinss 
- [x] Add icon to ;stats shard status - @JamesHawkinss 
- [x] Settings list improve - @JamesHawkinss 
- [x] setting set. Turn setting in code block - @JamesHawkinss 
- [x] Reformat flags to ICON - FLAG - @JamesHawkinss 
- [x] Add advertisement to flags command - @JamesHawkinss 
- [x] Tag needs alias "tags" - @JamesHawkinss 
- [x] Tag list needs reformat and not on usage - @JamesHawkinss 
- [x] ;tag shows up on usage - @JamesHawkinss 
- [x] ;tag `<tag>` needs improved description - @JamesHawkinss 
- [x] Inconsistent tag/name in usage - @JamesHawkinss 
- [x] Softban no args should reply usage - @JamesHawkinss 
- [x] Todo should have alias "todolist" - @JamesHawkinss 
- [x] Todo create > add > remove = missing embed. - @JamesHawkinss 
- [x] Inconsistent naming in todo usage - @JamesHawkinss 
- [x] usage missing `help [module]` - @JamesHawkinss 
- [x] ;loop invalid mode broken - @JamesHawkinss 
- [x] ;help `<invalid module>` broken - @JamesHawkinss 
- [ ] ;loop [mode] needs elaboration on modes - @JamesHawkinss 
- [x] Skip usage [force] needs to be sub command - @JamesHawkinss 
- [x] Queue usage [scope] needs to be sub command or extra description - @JamesHawkinss 

# Deadly
- [x] Lang string color rgb - @DeadlyFirex 
- [x] Color rgb accept spaces - @DeadlyFirex 
- [x] Settings no args: show list - @DeadlyFirex 
- [x] Random color language string missing - @DeadlyFirex 
- [x] Random command needs min max - @DeadlyFirex 
- [x] Random [amount] needs better description - @DeadlyFirex 
- [x] Module list not in usage - @DeadlyFirex 
- [x] Module list needs reformat - @DeadlyFirex 
- [x] Settings needs alias "setting" - @DeadlyFirex 
- [ ] Tag category delete everything in category (needs confirmation) - @DeadlyFirex 
- [x] Purge should ignore own message - @DeadlyFirex 
- [ ] Purge contains `<token>` not clear enough - @DeadlyFirex 
[...]

PR #191, opened 2019-11-01
Add a cancel button to the confirmation prompt

binaryoverload:
# Pull request

## Added/Changed
- Simple cancel button for any confirmation prompt

>  Forceban needs cancel reaction

From #189

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/utils/ConfirmUtils.java
+++ b/src/main/java/org/cascadebot/cascadebot/utils/ConfirmUtils.java
@@ -60,6 +60,10 @@ public class ConfirmUtils {
                     if (runner.getIdLong() != action.userId) return;
                     action.run();
                 }));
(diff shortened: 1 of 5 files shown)


src/main/java/org/cascadebot/cascadebot/utils/ConfirmUtils.java
```diff
@@ -60,6 +60,10 @@ public static boolean confirmAction(long userId, String actionKey, TextChannel c
                     if (runner.getIdLong() != action.userId) return;
                     action.run();
                 }));
+                group.addButton(new Button.UnicodeButton(UnicodeConstants.RED_CROSS, ((runner, channel1, message1) -> {
```
weeryan17: should this be a parameter in case we want a confirm that gets timed out?
@mrjvs what do you think?

PR #201, opened 2019-12-21
Joke command!

JamesHawkinss:
# Pull request

## Added/Changed
Added joke command
Changed WebUtils to always require json to be returned
Languified joke command


src/main/java/org/cascadebot/cascadebot/commands/fun/RandomJokeSubCommand.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public String command() {
+        return "joke";
+    }
+
+    @Override
+    public CascadePermission getPermission() {
+        return null;
```
weeryan17: their needs to be a thing here

JamesHawkinss:
> Why is this even a thing. also if we do want this to be a thing it should have a dedicated command, not be thrown in with random.

But it fits in nicely with the random command - it's returning a random joke each time, and I thought it worked better as a subcommand than as its own command.

If you want me to move it over to be its own command then I'd be more than happy to do that.

weeryan17:
> But it fits in nicely with the random command - it's returning a random joke each time, and I thought it worked better as a subcommand than as its own command.
> 
> If you want me to move it over to be its own command then I'd be more than happy to do that.

Please do so. this is the type of thing that could easily be want to be expanded in the future with different types of jokes and that stuff.

JamesHawkinss:
> > But it fits in nicely with the random command - it's returning a random joke each time, and I thought it worked better as a subcommand than as its own command.
> > If you want me to move it over to be its own command then I'd be more than happy to do that.
> 
> Please do so. this is the type of thing that could easily be want to be expanded in the future with different types of jokes and that stuff.

All done :)

src/main/java/org/cascadebot/cascadebot/commands/fun/JokeCommand.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    public void onCommand(Member sender, CommandContext context) {
+        try {
+            context.getTypedMessaging().replyInfo(getJoke());
+        } catch (IOException e) {
+            context.getTypedMessaging().replyDanger(context.i18n("commands.joke.error_loading"));
+        };
+    }
+
+    private String getJoke() throws IOException {
```
DeadlyFirex: In my opinion, rename this to `randomJoke()` and define it under the RandomUtils class, I don't see any point in having it here

src/main/java/org/cascadebot/cascadebot/commands/fun/JokeCommand.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class JokeCommand implements ICommandMain {
+
+    @Override
+    public void onCommand(Member sender, CommandContext context) {
+        try {
+            context.getTypedMessaging().replyInfo(getJoke());
+        } catch (IOException e) {
+            context.getTypedMessaging().replyDanger(context.i18n("commands.joke.error_loading"));
+        };
```
DeadlyFirex:
```suggestion
        }
```

DeadlyFirex: Unnecessary semicolon

src/main/resources/lang/en-GB.json
```diff
@@ -70,6 +70,11 @@
       "header_members": "No. Users",
       "header_color": "Colour"
     },
+    "joke": {
+      "command": "joke",
+      "description": "Returns a random joke.",
+      "error_loading": "Could not load joke!"
```
DeadlyFirex:
```suggestion
      "error_loading": "We were not able to load up a joke!"
```

src/main/java/org/cascadebot/cascadebot/utils/RandomUtils.java
```diff
@@ -38,4 +40,9 @@ public static String randomColor() {
         return result.toString();
     }
 
+    public String randomJoke() throws IOException {
+        JsonElement jsonElement = WebUtils.getJsonFromURL("https://icanhazdadjoke.com/");
```
binaryoverload: This can return a IllegalArgumentException, best to handle it

src/main/java/org/cascadebot/cascadebot/commands/fun/JokeCommand.java
```diff
@@ -0,0 +1,42 @@
...
+    @Override
+    public CascadePermission getPermission() {
+        return CascadePermission.of("joke", true);
+    }
+
+    @Override
+    public Module getModule() {
+        return Module.FUN;
+    }
+}
```
weeryan17: Blank line before end of class

JamesHawkinss: clarification pls

DeadlyFirex:
```suggestion
    public Module getModule() {
        return Module.FUN;
    }
    
}
```

PR #207, opened 2020-02-19
Music fixes

weeryan17:
# Pull request

## Added/Changed
Better music error handling
New audio track user data so I could do the better error handling
Added QueueMoveSubCommand
Added QueueRemoveSubCommand
Added MovableList for lists with items that the user can move.
Added the appropriate methods to cascade player for those command.
Did proper audio track user data checks as we didn't really have any checks on it before.

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/commands/music/QueueMoveSubCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/music/QueueMoveSubCommand.java
@@ -5,7 +5,6 @@
 
 package org.cascadebot.cascadebot.commands.music;
 
-import com.sedmelluq.discord.lavaplayer.track.AudioTrack;
 import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.entities.Member;
 import org.cascadebot.cascadebot.UnicodeConstants;
@@ -13,14 +12,14 @@ import org.cascadebot.cascadebot.commandmeta.CommandContext;
 import org.cascadebot.cascadebot.commandmeta.ISubCommand;
 import org.cascadebot.cascadebot.messaging.MessageType;
 import org.cascadebot.cascadebot.messaging.MessagingObjects;
+import org.cascadebot.cascadebot.music.MovableAudioTrack;
 import org.cascadebot.cascadebot.permissions.CascadePermission;
 import org.cascadebot.cascadebot.utils.buttons.Button;
 import org.cascadebot.cascadebot.utils.buttons.ButtonGroup;
+import org.cascadebot.cascadebot.utils.move.MovableList;
 
-import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
 
 public class QueueMoveSubCommand implements ISubCommand {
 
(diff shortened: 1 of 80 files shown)


JamesHawkinss:
You're sending a few messages in this commands, do they need to be languified?

weeryan17:
> You're sending a few messages in this commands, do they need to be languified?

Yegh i forgot to do language stuff. Give me a min. Thanks for reminding me.

src/main/java/org/cascadebot/cascadebot/utils/move/MovableList.java
```diff
@@ -0,0 +1,179 @@
...
+        int highestTrackDisplay = highestTrack;
+
+        if (lowestTrack < 0) {
+            highestTrackDisplay += Math.abs(lowestTrack);
+            lowestTrackDisplay = 0;
+        }
+
+        if (highestTrack >= list.size() - 1) {
+            lowestTrackDisplay -= highestTrack - (list.size() - 2);
+            highestTrackDisplay = list.size() - 2;
```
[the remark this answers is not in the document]
weeryan17: otherwise it would be out of the array if i remember right.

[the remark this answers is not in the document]
binaryoverload: *then comment it/store it in a variable* ;)

src/main/java/org/cascadebot/cascadebot/commands/music/QueueMoveSubCommand.java
```diff
@@ -0,0 +1,126 @@
...
+import org.cascadebot.cascadebot.utils.move.MovableList;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class QueueMoveSubCommand implements ISubCommand {
+
+    @Override
+    public void onCommand(Member sender, CommandContext context) {
+        if (context.getArgs().length >= 2) {
```
binaryoverload, line 28: This command only accepts two arguments or no arguments right? If so, we should send a usage if they enter >2 or 1 argument

src/main/java/org/cascadebot/cascadebot/commands/music/QueueRemoveSubCommand.java
```diff
@@ -0,0 +1,52 @@
...
+import org.cascadebot.cascadebot.commandmeta.CommandContext;
+import org.cascadebot.cascadebot.commandmeta.ISubCommand;
+import org.cascadebot.cascadebot.permissions.CascadePermission;
+
+public class QueueRemoveSubCommand implements ISubCommand {
+
+    @Override
+    public void onCommand(Member sender, CommandContext context) {
+        if (context.getArgs().length == 0) {
+            context.getUIMessaging().replyUsage(this);
```
binaryoverload, line 18:
```suggestion
            context.getUIMessaging().replyUsage();
```

src/main/java/org/cascadebot/cascadebot/data/objects/GuildData.java
```diff
@@ -68,6 +72,26 @@
 
     //endregion
 
+    //region Data Loaded Methods
+    public void guildLoaded() {
```
binaryoverload, line 76: Maybe rename this to be more like a event? i.e. `onGuildLoaded`

src/main/java/org/cascadebot/cascadebot/data/objects/GuildSettingsMusic.java
```diff
@@ -0,0 +1,33 @@
...
+@Getter
+@Setter
+@NoArgsConstructor(access = AccessLevel.PUBLIC)
+public class GuildSettingsMusic {
+
+    @Setting
+    private boolean preserveVolume = false;
+
+    @Setting
+    private boolean preserveEqualizer = false;
```
binaryoverload, line 27: I believe both of these options should be set to true by default imo

src/main/java/org/cascadebot/cascadebot/data/objects/GuildSettingsMusic.java
```diff
@@ -0,0 +1,33 @@
...
+import lombok.Setter;
+import org.cascadebot.cascadebot.commandmeta.Module;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@SettingsContainer(module = Module.MUSIC)
+@Getter
+@Setter
+@NoArgsConstructor(access = AccessLevel.PUBLIC)
```
binaryoverload, line 20: There's no need to make a public no-args constructor if you do not have any other constructors. It is included into Java by default. This can be removed safely :)

src/main/java/org/cascadebot/cascadebot/data/objects/GuildSettingsMusic.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Setting
+    private boolean preserveVolume = false;
+
+    @Setting
+    private boolean preserveEqualizer = false;
+
+    private int volume = 100;
+
+    private Map<Integer, Float> equalizerBands = new HashMap<>();
```
binaryoverload, line 31: Can we use a ConcurrentHashMap here please?

src/main/java/org/cascadebot/cascadebot/events/PlayerListener.java
```diff
@@ -56,10 +62,18 @@ private void onEnd(AudioTrack track) {
                 }
                 // Take the next track in the queue, remove it from the queue and play it
                 AudioTrack audioTrack = player.getQueue().remove();
-                player.playTrack(audioTrack);
+                try {
+                    player.playTrack(audioTrack);
+                } catch (FriendlyException e) {
```
binaryoverload, line 67: Have you tested that this actually works? I was under the impression that a different thread throws this so we have to use some sort of callback with Lavaplayer.

src/main/resources/arguments.json
```diff
@@ -191,6 +191,19 @@
           "_type": "optional"
         }
       }
+    },
+    "move*": {
+      "pos1*": {
+        "_type": "optional",
+        "pos2": {
```
binaryoverload, line 198: `pos1` -> `source`
`pos2` -> `destination`

Maybe?

src/main/java/org/cascadebot/cascadebot/messaging/MessagingUI.java
```diff
@@ -196,7 +199,7 @@ public void sendTracksFound(List<AudioTrack> tracks) {
         }
     }
 
-    public void checkPlaylistOrSong(String input, List<AudioTrack> tracks, CommandContext context) {
+    public void checkPlaylistOrSong(String input, List<AudioTrack> tracks, CommandContext context, boolean top) {
```
binaryoverload, line 202: It isn't quite clear what "top" means in this context. Perhaps either add a JavaDoc or name it something a little more descriptive?

src/main/java/org/cascadebot/cascadebot/data/objects/GuildSettingsMusic.java
weeryan17: I thought because both of these are donation stuffs they should be off by default. Granted I should really figure out how this needs to work. We don't want people who stop donating to keep being able to have a volume level.

src/main/java/org/cascadebot/cascadebot/events/PlayerListener.java
weeryan17: I don't know of a known bad song so I haven't been able to test this. I'll see if i can find one. Also I'm pretty sure the playTrack stuff is all one thread at least to the point where it would throw the exception, but I'll double check that.

src/main/java/org/cascadebot/cascadebot/data/objects/GuildSettingsMusic.java
binaryoverload: Hmm fair point, we'll discuss this :)

binaryoverload:
jenkins build

src/main/java/org/cascadebot/cascadebot/commands/music/QueueMoveSubCommand.java
weeryan17: No it accepts 1 argument as a start position for using the gui move, but I forgot to implement that.

binaryoverload: Ah okay, anyway, we should still throw a usage if it's > 2 length

src/main/resources/lang/en-GB.json
```diff
@@ -1119,7 +1119,10 @@
     "guild#flag#list": {
       "name": "list",
       "description": "Lists all flags."
-    }
+    },
+    "pos": "position",
+    "source": "source",
+    "destination": "destination"
```
binaryoverload, line 1125: Just a finicky thing but could this be above the other arguments for the sake of consistency ❤️

PR #209, opened 2020-03-28
Per guild permissions

weeryan17:
# Pull request

## Added/Changed
Changed most permissions checks over to per guild permission checks
Added per guild permission manager to function as the permission manager for a specific guild
Added PermissionObject interface with methods needed for making an object have a per guild permission.


src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
```diff
@@ -134,19 +140,19 @@ default void stop() {
         stopTrack();
     }
 
-    default void loadLink(String input, long requestUser, Consumer<String> noMatchConsumer, Consumer<FriendlyException> exceptionConsumer, Consumer<List<AudioTrack>> resultTracks) {
+    default void loadLink(String input, long requestUser, Consumer<String> noMatchConsumer, Consumer<FriendlyException> exceptionConsumer, Consumer<List<AudioTrack>> resultTracks, long channel, long guild) {
```
binaryoverload: Since this parameter list is getting quite long, it might be better to take requestUser, channel and guild and instead just pass in the TrackData.

src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
```diff
@@ -164,25 +170,25 @@ public void loadFailed(FriendlyException e) {
         });
     }
 
-    default void loadPlaylist(String name, Member sender, BiConsumer<LoadPlaylistResult, List<AudioTrack>> consumer) {
+    default void loadPlaylist(String name, Member sender, long channel, long guildId, BiConsumer<LoadPlaylistResult, List<AudioTrack>> consumer) {
```
binaryoverload: Same as above

src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
```diff
@@ -164,25 +170,25 @@ public void loadFailed(FriendlyException e) {
...
                 consumer.accept(LoadPlaylistResult.LOADED_USER, tracks);
-            });
+            }, channel, guildId);
         } else {
             consumer.accept(LoadPlaylistResult.DOESNT_EXIST, null);
         }
     }
 
-    default void loadPlaylist(String name, Member sender, PlaylistType scope, BiConsumer<LoadPlaylistResult, List<AudioTrack>> consumer) {
+    default void loadPlaylist(String name, Member sender, PlaylistType scope, long channel, long guild, BiConsumer<LoadPlaylistResult, List<AudioTrack>> consumer) {
```
binaryoverload: Same as above

src/main/java/org/cascadebot/cascadebot/music/CascadePlayer.java
```diff
@@ -204,10 +210,10 @@ default void loadPlaylist(String name, Member sender, PlaylistType scope, BiCons
         LoadPlaylistResult loadPlaylistResult = result;
         loadLoadedPlaylist(playlist, sender.getIdLong(), tracks -> {
             consumer.accept(loadPlaylistResult, tracks);
-        });
+        }, channel, guild);
     }
 
-    default void loadLoadedPlaylist(Playlist playlist, long reqUser, Consumer<List<AudioTrack>> loadedConsumer) {
+    default void loadLoadedPlaylist(Playlist playlist, long reqUser, Consumer<List<AudioTrack>> loadedConsumer, long channel, long guild) {
```
binaryoverload: Same as above

src/main/java/org/cascadebot/cascadebot/permissions/CascadePermission.java
```diff
@@ -27,7 +27,7 @@
     private final Module module;
 
 
-    private CascadePermission(String permissionRaw, boolean defaultPerm, Module module, Permission... discordPerm) {
+    public CascadePermission(String permissionRaw, boolean defaultPerm, Module module, Permission... discordPerm) {
```
binaryoverload: Why has this been changed? There are factory methods to create a CascadePermission

src/main/java/org/cascadebot/cascadebot/data/objects/PerGuildPermissionsManager.java
```diff
@@ -0,0 +1,108 @@
...
+public class PerGuildPermissionsManager {
+
+    private HashMap<String, GuildPermission> guildPermissions = new HashMap<>();
+
+    public PermissionsManager getCascadePermissionManager() {
+        return CascadeBot.INS.getPermissionsManager();
+    }
+
+    // Method that registers all hte permissions after the data was loaded
+    public void registerPermissionsDataLoaded(GuildData data) {
```
binaryoverload:
```suggestion
    public void registerPermissions(GuildData data) {
```

src/main/java/org/cascadebot/cascadebot/data/objects/GuildSettingsMusic.java
```diff
@@ -0,0 +1,33 @@
...
+@Getter
+@Setter
+@NoArgsConstructor(access = AccessLevel.PUBLIC)
+public class GuildSettingsMusic {
+
+    @Setting
+    private boolean preserveVolume = false;
+
+    @Setting
+    private boolean preserveEqualizer = false;
```
binaryoverload: I'd personally argue that these should be on by default

src/main/java/org/cascadebot/cascadebot/permissions/PermissionsManager.java
```diff
@@ -164,4 +156,8 @@ public SecurityLevel getUserSecurityLevel(long userId) {
         return Security.getLevelById(userId, DiscordUtils.getAllOfficialRoleIds(userId));
     }
 
+    public Map<String, CascadePermission> getPermissionMap() {
+        return permissions;
```
binaryoverload: This should using Map.copyOf to return an immutable copy :)

src/main/java/org/cascadebot/cascadebot/data/objects/PermissionObject.java
```diff
@@ -0,0 +1,24 @@
...
+    String getPermission();
+
+    String getParent();
+
+    Module cascadeModule();
+
+    default CascadePermission getInternalPermission() {
+        return CascadePermission.of(getParent() + "." + getPermission(), false, cascadeModule());
+    }
+}
```
binaryoverload, line 24: I think we should start to make things like these abstract classes if they have method bodies. I believe we've talked about this before.

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildData.kt
```diff
@@ -51,6 +50,9 @@ class GuildData(@field:Id val guildId: Long) {
 
     @Transient
     val pageCache = PageCache()
+
+    @Transient
+    val guildPermissionsManager = PerGuildPermissionsManager()
```
binaryoverload, line 55: Could we rename this to permissionsManager? Just because it's in guildData so I don't think we need another guild in the name

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildSettingsCore.kt
```diff
@@ -72,5 +74,20 @@ class GuildSettingsCore(guildId: Long) {
...
+
+    fun hasTag(key: String?): Boolean {
+        return tags.containsKey(key)
+    }
+
+    fun addTag(key: String?, tag: Tag?) {
+        tags[key!!] = tag!!
+    }
+
+    fun removeTag(key: String?): Boolean {
```
binaryoverload, line 89: All of these parameters should be non-nullable

PR #210, opened 2020-04-19
Kotlin

binaryoverload:
# Pull request

### What needs doing
 - [x] Making sure everything compiles in both Kotlin and Java
 - [x] Remove all duplicate classes
 - [x] Ensure that the MongoDB serialisation still works with Kotlin

### Potential Improvements
 - Make special reply methods an extension on the JDA channel


binaryoverload:
I have updated this so that everything seems to work in Kotlin now. I advise we get this merged as soon as possible because of the many many merge conflicts it could cause

src/main/kotlin/org/cascadebot/cascadebot/data/objects/PlaylistType.kt
```diff
@@ -0,0 +1,5 @@
+package org.cascadebot.cascadebot.data.objects
+
+enum class PlaylistType {
```
weeryan17, line 3: Why was this converted to a kotlin file?

binaryoverload: No reason for it, I was mainly figuring out if there was a difference.

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildCommandInfo.kt
```diff
@@ -0,0 +1,28 @@
+package org.cascadebot.cascadebot.data.objects
+
+import org.cascadebot.cascadebot.commandmeta.ICommandMain
+import org.cascadebot.cascadebot.data.language.Locale
+
+class GuildCommandInfo(var enabled: Boolean,
+                       command: String,
+                       defaultCommand: String,
+                       var aliases: MutableSet<String>) {
```
weeryan17, line 9: This class needs to be updated to the new decision to not allow changing of commands

src/main/kotlin/org/cascadebot/cascadebot/music/MusicHandler.kt
```diff
@@ -0,0 +1,197 @@
...
+        val twitchAudioManager = TwitchStreamAudioSourceManager()
+        twitchSourceName = twitchAudioManager.sourceName
+        playerManager.registerSourceManager(twitchAudioManager)
+        playerManager.registerSourceManager(BeamAudioSourceManager())
+        playerManager.registerSourceManager(SoundCloudAudioSourceManager.createDefault())
+        playerManager.registerSourceManager(BandcampAudioSourceManager())
+        if (Config.INS.musicNodes.size > 0) {
+            lavaLink = JdaLavalink(Config.INS.botId.toString(), Config.INS.shardNum, Function { shardId: Int? -> CascadeBot.INS.shardManager.getShardById(shardId!!) })
+            for (musicNode in Config.INS.musicNodes) {
+                lavaLink!!.addNode(musicNode.uri, musicNode.password) //TODO give nodes a name
```
weeryan17, line 73: I'm pretty sure lava link can never be null so the `!!` isn't needed

src/main/kotlin/org/cascadebot/cascadebot/music/MusicHandler.kt
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    /**
+     * Searches for a list of 5 tracks and if it errors send the error to the specified channel
+     *
+     * @param search  The string to search
+     * @param channel The [TextChannel] to send any errors to
+     * @return The list of tracks that where found
+     */
+    fun searchTracks(search: String?, channel: TextChannel, searchResultConsumer: Consumer<List<SearchResult>?>) {
```
weeryan17, line 120: `List<SearchResult>` is never null

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildCommandInfo.kt
binaryoverload: Yeah I half did that, I will finish that 👍

binaryoverload: @weeryan17 Done! :)

src/main/kotlin/org/cascadebot/cascadebot/music/MusicHandler.kt
```diff
@@ -0,0 +1,197 @@
...
+    val playerManager: AudioPlayerManager = DefaultAudioPlayerManager()
+    val typePattern = Pattern.compile("youtube#([A-z]+)")
+
+    val youtubeSourceName: String
+    val twitchSourceName: String
+
+    val players: MutableMap<Long, CascadePlayer> = HashMap()
+    var musicJsonParser = JsonParser()
+    lateinit var lavaLink: JdaLavalink
+        private set
```
weeryan17, line 52: wat

src/main/kotlin/org/cascadebot/cascadebot/utils/ChangeList.kt
```diff
@@ -0,0 +1,36 @@
+package org.cascadebot.cascadebot.utils
+
+import java.util.Collections
+
+class ChangeList<T>(val baseCollection: Set<T>) {
```
weeryan17, line 5: Why do we need this?

src/main/kotlin/org/cascadebot/cascadebot/music/MusicHandler.kt
binaryoverload: This is a variable that is late initialised and has a private setter. What's the problem?

src/main/kotlin/org/cascadebot/cascadebot/utils/ChangeList.kt
binaryoverload: Used for aliases in the command info, it's so that we can still update global aliases even if people have edited. I'd recommend checking the code out in command info

src/main/kotlin/org/cascadebot/cascadebot/music/MusicHandler.kt
weeryan17: ahhhh. the `private set` looked like a variable

src/main/kotlin/org/cascadebot/cascadebot/utils/ChangeList.kt
weeryan17: I see that it uses this, but don't see why this is used. Could you go through why we have a separate list class for handling this.

weeryan17: I see that it uses this, but don't see why this is used. Could you go through why we have a separate list class for handling this. From the looks of it its letting you add and remove items but apply it later but for what purpose. I just don't want us to have classes we don't need so I want to make sure we need this.

src/main/kotlin/org/cascadebot/cascadebot/utils/ChangeList.kt
```diff
@@ -0,0 +1,36 @@
...
+        get() = added.toList()
+    val removedItems
+        get() = removed.toList()
+
+    private val added = Collections.synchronizedSet<T>(HashSet())
+    private val removed = Collections.synchronizedSet<T>(HashSet())
+
+    fun add(element: T): Boolean {
+        if (!added.remove(element)) {
+            removed.add(element)
```
weeryan17, line 17: Yes add it to the remove list on add.

src/main/kotlin/org/cascadebot/cascadebot/utils/ChangeList.kt
```diff
@@ -0,0 +1,36 @@
...
+        }
+        return !baseCollection.contains(element)
+    }
+
+    fun remove(element: T): Boolean {
+        if (!removed.remove(element)) {
+            added.add(element)
```
weeryan17, line 24: Opposite of above

weeryan17:
Resolve conflicts and then we can merge

PR #213, opened 2020-04-25
Fix mention prefix due to mismatch between different mentions. Fixes #212

### LINKED ISSUE #212: [Bug] Mentioning cascade doesn't show prefix
**Describe the bug**
mentioning cascade doesnt show prefix

**To Reproduce**
1. mention cascade

**Expected behavior**
show prefix

weeryan17:
# Pull request

## Added/Changed
Fixed mention prefix due to having a mismatch between the two different ways of mentioning users

## Feature
N/A

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/events/CommandListener.java
+++ b/src/main/java/org/cascadebot/cascadebot/events/CommandListener.java
@@ -8,6 +8,7 @@ package org.cascadebot.cascadebot.events;
 import io.prometheus.client.Summary;
 import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.Permission;
+import net.dv8tion.jda.api.entities.Member;
 import net.dv8tion.jda.api.entities.MessageType;
 import net.dv8tion.jda.api.events.message.guild.GuildMessageReceivedEvent;
 import net.dv8tion.jda.api.hooks.ListenerAdapter;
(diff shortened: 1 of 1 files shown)


src/main/java/org/cascadebot/cascadebot/events/CommandListener.java
```diff
@@ -81,10 +82,17 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
         String trigger;
         String[] args;
 
+        boolean messagesBot = false;
+        for (Member member : event.getMessage().getMentionedMembers()) {
+            if (member.getIdLong() == event.getGuild().getSelfMember().getIdLong()) {
+                messagesBot = true;
+            }
+        }
+
```
binaryoverload, line 91: This won't work as intended as you could mention the bot at any point in the message.

We need to do what we did but just use maybe regex for it to match the extra character.

src/main/java/org/cascadebot/cascadebot/events/CommandListener.java
```diff
@@ -81,10 +82,17 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
...
         if (message.startsWith(prefix)) {
             commandWithArgs = message.substring(prefix.length()); // Remove prefix from command
-        } else if (guildData.getCoreSettings().isMentionPrefix() && message.startsWith(event.getJDA().getSelfUser().getAsMention())) {
-            commandWithArgs = message.substring(event.getJDA().getSelfUser().getAsMention().length()).trim();
+        } else if (guildData.getCoreSettings().isMentionPrefix() && messagesBot) {
+            commandWithArgs = message.substring(event.getJDA().getSelfUser().getAsMention().length() + 1).trim();
```
binaryoverload, line 95: I'm *pretty* sure that if we use context.getMember().getAsMention() this will work as before.

src/main/java/org/cascadebot/cascadebot/events/CommandListener.java
```diff
@@ -77,18 +80,32 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
...
         String trigger;
         String[] args;
 
+        // Check that the bot is even mentioned in the first place
+        boolean messagesBot = false;
+        for (Member member : event.getMessage().getMentionedMembers()) {
+            if (member.getIdLong() == event.getGuild().getSelfMember().getIdLong()) {
+                messagesBot = true;
+            }
+        }
```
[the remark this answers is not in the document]
weeryan17: I figured we want to check if the bot is mentioned at all before we check if its at the begining, yes it isn't needed as much with the regex, but still cuts out having to run the checking at start logic.

PR #215, opened 2020-05-03
Scheduled Actions

binaryoverload:
# Pull request

## Added/Changed
 - The concept of a "ScheduledAction" which is a predefined action that will be executed after a specified amount of time. 
 - Scheduled Actions will be used for things like reminders and temp moderation!
 - Scheduled Actions are stored in the "scheduled_actions" collection **immediately** after being registered. This collection is loaded at run time by the bot so that actions carry over bot-restarts.

**This PR does not contain the actions themselves, this is just the functionality to enable us to write actions!**
It does, however, contain two data structures for future data. This is mostly because I started to create the actions but realised this would hamper the usage of this feature on the dev branch. Hence, that is why a PR without actions is here :)


src/main/kotlin/org/cascadebot/cascadebot/scheduler/ScheduledAction.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    override fun run() {
+        try {
+            type.dataConsumer(this)
+        } finally {
+            ScheduledActionManager.deleteScheduledAction(this.id)
+        }
+    }
+
+    enum class ActionType(val expectedClass: KClass<*>, val dataConsumer: (ScheduledAction) -> Unit) {
```
weeryan17, line 56: Shouldn't this be in the enum file?

weeryan17, line 56: Why does this need a class if its just running a method?

src/main/kotlin/org/cascadebot/cascadebot/scheduler/ScheduledAction.kt
```diff
@@ -0,0 +1,78 @@
...
+        val userId: Long,
+        val creationTime: Instant,
+        val executionTime: Instant
+) : Runnable {
+
+    @Id
+    val id: ObjectId = ObjectId()
+
+    val delay: Long
+        get() = ChronoUnit.MILLIS.between(creationTime, executionTime)
```
weeryan17, line 25: This will cause issues when the node is shut down as its not calculating the delay between now and execution, its doing it between creation and execution. Also this is being stored in mongo when it should be calculated on load.

PR #218, opened 2020-05-04
Paged usage

weeryan17:
# Pull request

## Added/Changed
 - Added spiting usage when its over 1000 chars so we don't get issues with large usages.
 - Added option to have extra pages in the usage for describing the functions of a command if additional information is needed.
 - Improved page system
     - Stooped showing page number when only one page
     - Removed and add buttons based on current page
     - Have no last/first page button when only two pages

## Feature
N/A


src/main/kotlin/org/cascadebot/cascadebot/messaging/Messaging.kt
```diff
@@ -113,6 +114,68 @@ object Messaging {
...
+            group?.removeButton(lastPageButton)
+            if (pageGroup.pages > 2) {
+                group?.addButton(firstPageButton)
+            }
+            group?.addButton(prevPageButton)
+            group?.addButton(nextPageButton)
+            if (pageGroup.pages > 2) {
+                group?.addButton(lastPageButton)
+            }
+            pageGroup.isMissingButtons = false
```
binaryoverload, line 175: I'm unsure about this. For two reasons: 
 - The usage of it is fairly clunky when going from page 1 -> 2 and last page -> last page - 1
 - This is a *lot* of rest actions to be doing. I'm not sure it's worth the tiny (imo) UI improvement

src/main/kotlin/org/cascadebot/cascadebot/commandmeta/CommandContext.kt
```diff
@@ -141,6 +143,10 @@ class CommandContext(
         }
     }
 
+    fun replyUsage() {
+        uiMessaging.replyUsage()
+    }
+
```
binaryoverload, line 149: Just a method that proxies another message?

src/main/java/org/cascadebot/cascadebot/utils/pagination/PageCache.java
```diff
@@ -41,6 +43,13 @@ public int getPages() {
             return pages.size();
         }
 
+        public boolean isMissingButtons() {
+            return missingButtons;
```
binaryoverload, line 47: What is this used for?

src/main/kotlin/org/cascadebot/cascadebot/commandmeta/CommandContext.kt
weeryan17: Thanks for pointing this out. I thought more used the other usage method in context so I added this as a replacement, but not a lot of other do.

src/main/java/org/cascadebot/cascadebot/utils/pagination/PageCache.java
weeryan17: If the pages is missing buttons due to it being on the first page, or on the last.

src/main/kotlin/org/cascadebot/cascadebot/messaging/Messaging.kt
weeryan17: jvs told me he wanted pages to not have buttons when they shouldn't need them and that i could only do paged usage if i implemented those changes. And I'm fairly sure he'd complain if we didn't reset the buttons. I'll get him input on this though.
@mrjvs Whats your thoughts on this?

PR #226, opened 2020-06-09
Join a channel when a user plays a song

JamesHawkinss:
# Pull request

## Added/Changed
Join a channel when a user plays a song etc

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/commands/music/PlayCommand.java
+++ b/src/main/java/org/cascadebot/cascadebot/commands/music/PlayCommand.java
@@ -6,6 +6,7 @@
 package org.cascadebot.cascadebot.commands.music;
 
 import net.dv8tion.jda.api.entities.Member;
+import net.dv8tion.jda.api.entities.VoiceChannel;
 import org.cascadebot.cascadebot.CascadeBot;
 import org.cascadebot.cascadebot.commandmeta.CommandContext;
 import org.cascadebot.cascadebot.commandmeta.MainCommand;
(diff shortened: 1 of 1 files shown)


src/main/java/org/cascadebot/cascadebot/commands/music/PlayCommand.java
```diff
@@ -17,6 +18,16 @@
...
     public void onCommand(Member sender, CommandContext context) {
+        VoiceChannel voiceChannel = sender.getVoiceState().getChannel();
+
+        if (voiceChannel != null) {
+            if (context.getMusicPlayer().getConnectedChannel() != voiceChannel) {
+                context.runOtherCommand("join", sender, context);
+            }
+        } else {
+            context.runOtherCommand("join", sender, context);
+        }
```
binaryoverload, line 29: We need to add a setting to enable this behaviour. Maybe a `joinonplay` in MusicSettings

JamesHawkinss: Just committed. Did I do it right?

src/main/java/org/cascadebot/cascadebot/commands/music/PlayCommand.java
```diff
@@ -17,6 +18,18 @@
...
+        VoiceChannel voiceChannel = sender.getVoiceState().getChannel();
+
+        if (context.getData().getMusic().getJoinOnPlay()) {
+            if (voiceChannel != null) {
+                if (context.getMusicPlayer().getConnectedChannel() != voiceChannel) {
+                    context.runOtherCommand("join", sender, context);
+                }
+            } else {
+                context.runOtherCommand("join", sender, context);
+            }
```
binaryoverload, line 30: We can just ignore this (remove the else)

src/main/java/org/cascadebot/cascadebot/commands/music/PlayCommand.java
```diff
@@ -17,6 +18,18 @@
 
     @Override
     public void onCommand(Member sender, CommandContext context) {
+        VoiceChannel voiceChannel = sender.getVoiceState().getChannel();
+
+        if (context.getData().getMusic().getJoinOnPlay()) {
```
binaryoverload, line 23:
```suggestion
        if (context.getData().getMusic().getJoinOnPlay()) {
            VoiceChannel voiceChannel = sender.getVoiceState().getChannel();
```

src/main/java/org/cascadebot/cascadebot/commands/music/PlayCommand.java
weeryan17, line 30: Few things.
1. check if bot is already connected to something, if so don't run this code
2. We don't need to check if the user is in a vc at all. The play command already does that.

Issue #221, opened 2020-05-17
[Feature request] Vote skip takes a long time


mrjvs:
**Is your feature request related to a problem? Please describe.**
Skip vote takes a long time.

**Describe the solution you'd like**
Skip vote time cut by half. so 15 seconds.
30 seconds is a big portion of a song already. better of not skipping it al together if it takes a long time.
On top of that I would also like for it to check for if the more than 50% of people in vc has voted yes.
if thats the case then just skip.

**Describe alternatives you've considered**
Not applicable

**Additional context**
agreed by 2 other people in cascade guild
(dont count bots in the list of people in vc)

weeryan17:
I don't fully agree with lowering the vote time to the degree you suggested. I see the point in we might want to lower it, but we want to give people enough time to make sure everyone votes. This might be a thing we make a vote for our testers to see what they think.

JamesHawkinss:
I agree that the time should be shortened. Perhaps we can make it skip as soon as the threshold is met?

weeryan17:
![Screenshot from 2020-05-17 09-34-28](https://user-images.githubusercontent.com/10491247/82151570-a5ba1580-9821-11ea-976c-50f8e20dfeb7.png)

villChurch:
If this is still up for grabs then happy to have a go.

weeryan17:
Good luck. If you need any help let us know.

mrjvs:
The functionality (10s initial, +5s for every new vote) will probably require editing of the votes package (utils/votes). it only allows a static time without a way to update a time.

villChurch:
With this feature, I take it if 50% of people in vc voted you still want it to skip even with the new timer changes?

binaryoverload:
Yup! :)

PR #231, opened 2020-06-20
Vote to skip feature

villChurch:
Added functionality where vote to skip timer starts at 10 seconds and increases by 5 seconds everytime a new member votes, with a max of 30 seconds.

Also if at least 50% of the voice channel vote it will skip straight away.

# Pull request

 - [X] I have read and agreed to the [code of conduct](https://github.com/CascadeBot/CascadeBot/blob/master/.github/CODE_OF_CONDUCT.md).
 - [X] I have read and complied with the [contributing guidelines](https://github.com/CascadeBot/CascadeBot/blob/master/.github/CONTRIBUTING.md).
 - [X] What I'm implementing was assigned to me or was not being worked on and is a somewhat requested feature. For reference see our [GitHub projects](https://github.com/orgs/CascadeBot/projects).
 - [X] I have tested all of my changes.
 
## Added/Changed
What did you add and/or change for this.
Changes to the VoteButtonGroup and VoteButtonGroupBuilder in side the utils/votes package.

## Feature
#221

### DIFF
--- a/src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
+++ b/src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
@@ -142,7 +142,7 @@ public class VoteButtonGroup extends PersistentButtonGroup {
 
             int votesNeeded = (int) Math.ceil(member.getVoiceState().getChannel().getMembers().size() / 2);
 
-            if ((votes.size() - 1) >= votesNeeded) { // - 1 is because the users vote hasn't been added yet
+            if ((votes.size()) >= votesNeeded - 1) { // - 1 is because the users vote hasn't been added yet
                 CascadeBot.INS.getShardManager().getGuildById(getGuildId()).getTextChannelById(getChannelId()).retrieveMessageById(getMessageId()).queue(message -> {
                     message.delete().queue(null, DiscordUtils.handleExpectedErrors(ErrorResponse.UNKNOWN_MESSAGE));
                     voteFinished();
--- a/src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
+++ b/src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
@@ -6,7 +6,6 @@
 package org.cascadebot.cascadebot.utils.votes;
 
 import com.google.common.collect.Sets;
-import net.dv8tion.jda.api.entities.Guild;
 import net.dv8tion.jda.api.entities.Member;
 import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.entities.User;
@@ -16,8 +15,6 @@ import org.cascadebot.cascadebot.utils.DiscordUtils;
 import org.cascadebot.cascadebot.utils.FormatUtils;
 import org.cascadebot.cascadebot.utils.buttons.PersistentButtonGroup;
 
-import java.math.RoundingMode;
-import java.text.DecimalFormat;
 import java.time.Instant;
 import java.util.Collections;
 import java.util.HashMap;
@@ -45,19 +42,43 @@ public class VoteButtonGroup extends PersistentButtonGroup {
 
     private Timer voteTimer;
 
-    private long timerStartTime;
+    private long timerStartTime = Instant.now().toEpochMilli();
 
-    private int timerRunTime = 10;
+    private int timerRunTime;
 
-    VoteButtonGroup(long ownerId, long channelId, long guildId, BiConsumer<List<VoteResult>, Message> periodicRunnable, Timer voteTimer, long timerStartTime, Consumer<List<VoteResult>> finishConsumer) {
+    private int maxTimerRunTime;
+
+    private int timerRunTimeSkipAddon;
+
+    private boolean isMusicSkip;
+
+    public void setIsMusicSkip(Boolean isMusicSkip) {
+        this.isMusicSkip = isMusicSkip;
+    }
+
+    public void setFinishConsumer(Consumer<List<VoteResult>> finishConsumer) {
+        this.finishConsumer = finishConsumer;
+    }
+
+    public void setMaxTimeRunTime(int maxTimerRunTime) {
+        this.maxTimerRunTime = maxTimerRunTime;
+    }
+
+    public void setTimerRunTimeSkipAddon(int timerRunTimeSkipAddon) {
+        this.timerRunTimeSkipAddon = timerRunTimeSkipAddon;
+    }
+
+    public void setTimerRunTime(int timerRunTime) {
+        this.timerRunTime = timerRunTime;
+    }
+
+    VoteButtonGroup(long ownerId, long channelId, long guildId, BiConsumer<List<VoteResult>, Message> periodicRunnable, Timer voteTimer) {
         super(ownerId, channelId, guildId);
         this.periodicConsumer = periodicRunnable;
         if (periodicRunnable != null) {
             setUpVoteProcessConsumer();
         }
         this.voteTimer = voteTimer;
-        this.timerStartTime = timerStartTime;
-        this.finishConsumer = finishConsumer;
     }
 
     private void setUpVoteProcessConsumer() {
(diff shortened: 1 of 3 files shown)


src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
```diff
@@ -65,6 +82,55 @@ public void addVote(User user, Object vote) {
                 votes.remove(user.getIdLong());
                 return;
             }
+        } else {
+            DecimalFormat df = new DecimalFormat("#");
+            df.setRoundingMode(RoundingMode.HALF_UP);
+            long timeElapsed = Instant.now().toEpochMilli() - timerStartTime;
+            int elapsed = Integer.parseInt(df.format(timeElapsed /1000));
```
binaryoverload, line 89: Instead of doing this to round, you can use our `FormatUtils#round`

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
```diff
@@ -65,6 +82,55 @@ public void addVote(User user, Object vote) {
...
+                voteTimer.cancel();
+                voteTimer = new Timer();
+                voteTimer.schedule(new TimerTask() {
+                    @Override
+                    public void run() {
+                        CascadeBot.INS.getShardManager().getGuildById(getGuildId()).getTextChannelById(getChannelId()).retrieveMessageById(getMessageId()).queue(message -> {
+                            message.delete().queue(null, DiscordUtils.handleExpectedErrors(ErrorResponse.UNKNOWN_MESSAGE));
+                        }, DiscordUtils.handleExpectedErrors(ErrorResponse.UNKNOWN_MESSAGE));
+                        voteFinished();
+                        finishConsumer.accept(getOrderedVoteResults());
```
binaryoverload, line 110: You should do both of these within the `retrieveMessageById` success callback

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
```diff
@@ -65,6 +82,55 @@ public void addVote(User user, Object vote) {
...
+                }, TimeUnit.SECONDS.toMillis(newTimersTime));
+
+                List<Guild> mutualGuilds = user.getMutualGuilds();
+                int position = 0;
+                for (int i = 0; i < mutualGuilds.size(); i++){
+                    if (mutualGuilds.get(i).getIdLong() == getGuildId()){
+                        position = i;
+                    }
+                }
+                Member member = mutualGuilds.get(position).getMember(user);
```
binaryoverload, line 121: You could do `CascadeBot.INS.getShardManager().getGuildById(getGuildId()).getMember(user)`

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
```diff
@@ -65,6 +82,55 @@ public void addVote(User user, Object vote) {
...
+
+                int votesNeeded = Integer.parseInt(df.format(((member.getVoiceState().getChannel().getMembers().size()) / 2)));
```
binaryoverload, line 123: A `Math.ceil` could be used here

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
```diff
@@ -65,6 +82,55 @@ public void addVote(User user, Object vote) {
...
+
+                if (votes.size() >= votesNeeded) {
+                    CascadeBot.INS.getShardManager().getGuildById(getGuildId()).getTextChannelById(getChannelId()).retrieveMessageById(getMessageId()).queue(message -> {
+                        message.delete().queue(null, DiscordUtils.handleExpectedErrors(ErrorResponse.UNKNOWN_MESSAGE));
+                    }, DiscordUtils.handleExpectedErrors(ErrorResponse.UNKNOWN_MESSAGE));
+                    voteFinished();
+                    finishConsumer.accept(getOrderedVoteResults());
+                    stopVote();
```
binaryoverload, line 131: Same as before, these should be run in the success consumer

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
villChurch: Would `Math.round` not be better to then cast to an int as `Math.ceil` of say 3 would output 2.0 whereas `Math.round` of 3 would output 2?

villChurch:
> This is a really solid PR! Just looking for some changes to make the code a little nicer :)

Thanks, just noticed it would stop looking to see if more than half the members in voice had voted if the max time limit was reached so i've made a change to rectify that, just moving an if clause, and made the suggested changes.

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
```diff
@@ -27,19 +37,27 @@
...
     private Timer timer = new Timer();
 
     private Timer voteTimer;
 
-    VoteButtonGroup(long ownerId, long channelId, long guildId, BiConsumer<List<VoteResult>, Message> periodicRunnable, Timer voteTimer) {
+    private long timerStartTime;
+
+    private int timerRunTime = 10;
+
+    VoteButtonGroup(long ownerId, long channelId, long guildId, BiConsumer<List<VoteResult>, Message> periodicRunnable, Timer voteTimer, long timerStartTime, Consumer<List<VoteResult>> finishConsumer) {
```
weeryan17, line 52: Why is the timer start an option? I could see it maybe as a separate method in case you have it in the future, but by default it should just be the system time, or the time the message is actuality sent (there's a consumer for that on button groups)

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
binaryoverload: The difference is that Math.ceil **always** rounds up. Which in this circumstance, I believe we'd want

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
villChurch: I passed it through so it was the exact time it started but you're right this can be replaced by just system time

villChurch:
@weeryan17 Is this changed how you wanted?

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroupBuilder.java
```diff
@@ -130,6 +136,13 @@ public VoteButtonGroupBuilder setPeriodicConsumer(BiConsumer<List<VoteResult>, M
      */
     public VoteButtonGroup build(long owner, long channelId, long guild) {
         VoteButtonGroup buttonGroup = new VoteButtonGroup(owner, channelId, guild, periodicConsumer, timer);
+        buttonGroup.setIsMusicSkip(isMusicSkip);
+        if (isMusicSkip) {
+            buttonGroup.setFinishConsumer(finishConsumer);
+            buttonGroup.setTimerRunTime(10);
+            buttonGroup.setMaxTimeRunTime(30);
+            buttonGroup.setTimerRunTimeSkipAddon(5);
```
weeryan17, line 144: Having these in this method will cause this method to eventually be way to big. These methods should be called in each individual instance (so in this case on the skip command its self).

binaryoverload:
@villChurch Hey Will, Just a small request from Ryan and then we can get this through! Let us know if you have any difficulties :)

src/main/java/org/cascadebot/cascadebot/utils/votes/VoteButtonGroup.java
```diff
@@ -27,12 +34,44 @@
...
 
+    private long timerStartTime = Instant.now().toEpochMilli();
+
+    private int timerRunTime;
+
+    private int maxTimerRunTime;
+
+    private int timerRunTimeSkipAddon;
+
+    private boolean isMusicSkip;
```
weeryan17, line 53: This needs a better name as this won't just be used for music

binaryoverload (approved):

PR #236, opened 2020-08-02
⏱ Temporary Moderation commands + Mute

binaryoverload:
# Pull request

## Added/Changed
 - Temp mute, temp ban, mute and unmute commands


binaryoverload:
jenkins build

src/main/kotlin/org/cascadebot/cascadebot/commands/moderation/MuteChannelSetupSubCommand.kt
```diff
@@ -0,0 +1,121 @@
...
+                futures.add(when (it) {
+                    is TextChannel -> it.manager.putPermissionOverride(mutedRole, 0, Permission.MESSAGE_WRITE.rawValue).submit()
+
+                    is Category -> {
+                        // Don't modify voice only categories
+                        if (it.textChannels.isEmpty()) return@forEach
+                        it.manager.putPermissionOverride(mutedRole, 0, Permission.MESSAGE_WRITE.rawValue).submit()
+                    }
+
+                    else -> return@forEach
```
weeryan17, line 79: We should also remove speak permissions in vcs for the user and/or server mute them. I don't see ether of those being done right now.

binaryoverload: Hmm 🤔 Not sure how I feel about the `;mute` command affecting speaking.

weeryan17: I personally would expect the mute command to also mute then in voice, and I'm sure other people would as well.

Ikinon: Also remember you can't server mute people if they are not in voice, you'll recieve bad request. Same goes with unmuting.

binaryoverload: I expected that might be the case, thanks Ikinon! :)

src/main/kotlin/org/cascadebot/cascadebot/commands/moderation/UnMuteCommand.kt
```diff
@@ -0,0 +1,57 @@
...
+
+        val mutedRole = context.guild.getMutedRole()
+
+        if (targetMember.roles.contains(mutedRole)) {
+            ScheduledActionManager.removeIf {
+                if (it.type != ActionType.UNMUTE) return@removeIf false
+                if (it.data !is ScheduledAction.ModerationActionData) return@removeIf false
+                it.data.targetId == targetMember.idLong
+            }
+        } else {
```
Ikinon, line 44: This seems as if we intend to unmute the member, but then we also don't reply when we have done it
We also don't include the reason to the audit log
Also we don't catch the error if we don't have permission to unmute the user

When adding reason to audit log, we just don't. It adds "{author} muted user", doesn't add reason. (This is across all of mute commands)

weeryan17: We can't send a reason to the audit logs. It's a role add, they don't have a reason attached to them (well they kind of do, but not in a way that matters).
Also I'm fairly sure the moderation code handles the replying. Could be wrong, I can't look right now as I'm on mobile.

binaryoverload: Ryan is right in that we can't actually log to audit log (But I will still add a reason in case this is added on Discord's end)

Ikiinon is right that this doesn't actually mute 😅 I obviously got too excited over my removeIf statement 😆

Ikinon: Don't understand what you mean in that you can't send a reason to audit logs in adding roles. If I'm understanding correctly you can, and in fact we already do this when we mute/unmute in temp mute, as so: https://i.imgur.com/pN9kZF3.png
My statement was about that we don't append the reason into this. I'd expect it to be {author} muted user: {reason}
If we are not adding the reason to the log, then other than maybe modlog, why have a reason parameter?

weeryan17: Mod log. Also I had no idea discord actually displayed that. I thought it just eat it. I'll look more into this when I get home.

src/main/kotlin/org/cascadebot/cascadebot/commands/moderation/MuteChannelSetupSubCommand.kt
```diff
@@ -0,0 +1,120 @@
...
+        context.guild.channels.forEach {
+            try {
+                futures.add(when (it) {
+                    is TextChannel -> it.manager.putPermissionOverride(mutedRole, setOf(), setOf(Permission.MESSAGE_WRITE)).submit()
+                    is VoiceChannel -> it.manager.putPermissionOverride(mutedRole, setOf(), setOf(Permission.VOICE_SPEAK)).submit()
+                    is Category -> {
+                        it.manager.putPermissionOverride(mutedRole, setOf(), setOf(Permission.MESSAGE_WRITE, Permission.VOICE_SPEAK)).submit()
+                    }
+
+                    else -> return@forEach
```
weeryan17, line 78: Have you tested if this respects channels synced with the category in terms of permissions? As in this doesn't unsync it due to setting the individual channel permissions. If it does unsync it we'll need to add check so we don't unsync it.

binaryoverload: Funnily enough, I found out the "channel sync" is actually handled in an interesting way. A channel is considered synced if all of it's permissions are the same as the category.

weeryan17: All so the client must do the syncing then. Alright then this is fine.

PR #238, opened 2020-08-09
🤔 Remindme command

### LINKED ISSUE #161: Some fun commands
- [x] color (hex, random + html colors)
- [x] random number
- [ ] remindme

binaryoverload:
# Pull request

## Added/Changed
 - The `;remindme` command to remind users of something

Closes #161

### DIFF
--- a/src/main/kotlin/org/cascadebot/cascadebot/commands/useful/RemindMeCommand.kt
+++ b/src/main/kotlin/org/cascadebot/cascadebot/commands/useful/RemindMeCommand.kt
@@ -10,6 +10,7 @@ import net.dv8tion.jda.api.entities.Member
 import org.cascadebot.cascadebot.commandmeta.CommandContext
 import org.cascadebot.cascadebot.commandmeta.MainCommand
 import org.cascadebot.cascadebot.commandmeta.Module
+import org.cascadebot.cascadebot.commandmeta.SubCommand
 import org.cascadebot.cascadebot.data.managers.ScheduledActionManager
 import org.cascadebot.cascadebot.permissions.CascadePermission
 import org.cascadebot.cascadebot.scheduler.ActionType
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/cascadebot/cascadebot/commands/useful/RemindMeCommand.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.cascadebot.cascadebot.utils.FormatUtils
+import org.cascadebot.cascadebot.utils.ParserUtils
+import java.time.Duration
+import java.time.Instant
+import java.time.OffsetDateTime
+
+class RemindMeCommand : MainCommand() {
+
+    override fun onCommand(sender: Member, context: CommandContext) {
+        val dm = context.testForArg("dm")
```
weeryan17, line 27: we should probably check to see if the bot can dm someone when this command is run.

src/main/resources/arguments.json
```diff
@@ -510,6 +510,23 @@
...
+    },
+    "dm*": {
+      "delay*": {
+        "_type": "required",
+        "message": {
+          "_type": "required"
+        }
+      }
+    },
+    "list": {}
```
weeryan17, line 528: this should have a type of `command`

binaryoverload: Doesn't technically need it, I tend to leave it blank

src/main/kotlin/org/cascadebot/cascadebot/commands/useful/RemindMeCommand.kt
binaryoverload: The only way we can actually do that iirc is to actually DM them and see if an error is thrown. I'll check

weeryan17: It should error on the opening of the channel afaik, so we can just open the channel to check.

binaryoverload: Yeah sounds good, I'll do that 👍

binaryoverload: Turns out this doesn't work as we thought it did @weeryan17! It only throws an error when sending the actual message.

I see 3 potential options here:
 - Ignore the error (Could cause confusion)
 - Send an error in the text-channel for every time a DM reminder is triggered that fails
 - Send an error in the text-channel and cancel the other reminders

weeryan17: I'd go with option 2.
On a side note is it possible to let people friend the bot so the bot can send dms to them via that method if they don't want to enable dms?

Ikinon: Bots can't friend, you'll get forbidden error as they can't use the endpoint.
I'd also go with option 2, also give the option on the message to attempt to re-send the DM, or to send the message in the channel.

PR #241, opened 2020-09-04
feature\Lock commands

Ikinon:
# Pull request

 - [ ] I have read and agreed to the [code of conduct](https://github.com/CascadeBot/CascadeBot/blob/master/.github/CODE_OF_CONDUCT.md).
 - [ ] I have read and complied with the [contributing guidelines](https://github.com/CascadeBot/CascadeBot/blob/master/.github/CONTRIBUTING.md).
 - [ ] What I'm implementing was assigned to me or was not being worked on and is a somewhat requested feature. For reference see our [GitHub projects](https://github.com/orgs/CascadeBot/projects).
 
## Feature
Was requested


binaryoverload (approved):

src/main/kotlin/org/cascadebot/cascadebot/commands/moderation/UnlockCommand.kt
```diff
@@ -0,0 +1,73 @@
...
+import net.dv8tion.jda.api.entities.TextChannel
+import net.dv8tion.jda.api.exceptions.PermissionException
+import org.cascadebot.cascadebot.commandmeta.CommandContext
+import org.cascadebot.cascadebot.commandmeta.MainCommand
+import org.cascadebot.cascadebot.commandmeta.Module
+import org.cascadebot.cascadebot.data.managers.LockManager
+import org.cascadebot.cascadebot.permissions.CascadePermission
+import org.cascadebot.cascadebot.utils.DiscordUtils
+
+class UnlockCommand : MainCommand() {
```
weeryan17, line 17: We might want to do a thing where we store when they locked a channel and if it's been a while then we warn the user that it will reset permissions to before the channel was lock, either that or we find a way to handle permissions being changed while a channel is locked.

weeryan17 (changes requested):

src/main/kotlin/org/cascadebot/cascadebot/commands/moderation/UnlockCommand.kt
binaryoverload: I don't think we need to do this since our lock command only adds 1 permission override per "lock"

weeryan17: Yegh my bad, for some reason I thought it was storing all permissions before the back, but apon further investigation it's not.

weeryan17 (approved):

PR #225, opened 2020-06-06
Feature/modlog

weeryan17:
# Pull request
For now I'm making this pr to get feedback on the code/system. I still need to do language, and a few minor TODOs as well as cascade related events. Also command have been only somewhat tested, and the events themselves have not been.

 - [x] I have read and agreed to the [code of conduct](https://github.com/CascadeBot/CascadeBot/blob/master/.github/CODE_OF_CONDUCT.md).
 - [x] I have read and complied with the [contributing guidelines](https://github.com/CascadeBot/CascadeBot/blob/master/.github/CONTRIBUTING.md).
 - [x] What I'm implementing was assigned to me or was not being worked on and is a somewhat requested feature. For reference see our [GitHub projects](https://github.com/orgs/CascadeBot/projects).
 - [ ] I have tested all of my changes.
 
## Added/Changed
 - [x] Modlog sending
 - [x] Modlog guild data
 - [ ] Storing modlog events
     - [x] Storing the actual events
     - [ ] Expire (will do on donation stuff later)
 - [x] Discord related events
 - [ ] Cascade related events
 - [ ] Message handling
     - [x] Storing of messages in redis
     - [ ] Encrypting messages with config defined key, and iv based off of message id (need to do iv)
     - [ ] Show extended message info (Need to talk to jvs about how exactly to handle)

PR #254, opened 2021-05-29
Feature/guilddata

weeryan17:
# Pull request 
Requires #225 

## Added/Changed
added DiffUtils in order to get the difference between arbitrary objects
Implemented new write system for guild data that we need to now switch over to any time we plan on writing to guild data
(Still need to do this)
Implemented listening to change steams in order to update guild data
Made it so @mrjvs has something to work on now 👀


src/main/java/org/cascadebot/cascadebot/CascadeBot.java
```diff
@@ -196,6 +198,17 @@ private void init() {
...
 
+        databaseManager.runAsyncTask(database -> {
+            ChangeStreamIterable<GuildData> changeStreamIterable = database.getCollection(GuildDataManager.COLLECTION, GuildData.class).watch();
+            new Thread(() -> changeStreamIterable.forEach(guildDataChangeStreamDocument ->
+                    GuildDataManager.replaceInternal(guildDataChangeStreamDocument.getFullDocument()), (result, throwable) -> {
+                if (throwable != null) {
+                    throwable.printStackTrace();
+                }
+                System.out.println(result);
+            }), "mongoChangeWatch").start();
```
binaryoverload, line 209: Do we need to start this in a separate thread? Since it's a callback I would have assumed it was non-blocking to be honest

src/main/java/org/cascadebot/cascadebot/data/managers/GuildDataManager.java
```diff
@@ -7,23 +7,35 @@
...
 
 public final class GuildDataManager {
 
-    private static final String COLLECTION = "guilds";
+    public static final String COLLECTION = "guilds";
 
     private static LoadingCache<Long, GuildData> guilds = Caffeine.newBuilder()
             .expireAfterAccess(5, TimeUnit.MINUTES)
-            .removalListener(new GuildSaveListener())
+            //.removalListener(new GuildSaveListener()) TODO do we want to use still use this?
```
binaryoverload, line 38: With the new system, there should be no reason to use this.

src/main/java/org/cascadebot/cascadebot/utils/diff/DiffUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    public static Difference diff(Object oldObj, Object newObj) {
+        return diff(new Difference(), "", oldObj, newObj);
+    }
+
+    @SneakyThrows
+    private static Difference diff(Difference currentDiff, String path, Object oldObj, Object newObj) {
+        Type mapType = new TypeToken<Map<String, Object>>() {
+        }.getType();
+        Map<String, Object> oldMap = CascadeBot.getGSON().fromJson(CascadeBot.getGSON().toJson(oldObj), mapType);
+        Map<String, Object> newMap = CascadeBot.getGSON().fromJson(CascadeBot.getGSON().toJson(newObj), mapType);
```
binaryoverload, line 28: Why are we using JSON here instead of reflection?

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildData.kt
```diff
@@ -44,12 +68,15 @@ class GuildData(@field:Id val guildId: Long) {
 
     //region Transient fields
     @Transient
+    @kotlin.jvm.Transient
```
binaryoverload, line 71: Why are these being added to everything? Does the normal Transient annotation not work?

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildData.kt
```diff
@@ -1,20 +1,44 @@
...
-class GuildData(@field:Id val guildId: Long) {
+class GuildData(@field:Id val guildId: Long): Cloneable {
+
+    @Transient
+    @kotlin.jvm.Transient
+    var lock: ReadWriteLock = ReentrantReadWriteLock()
+
+    @Transient
+    @kotlin.jvm.Transient
+    var writeMode = false;
```
binaryoverload, line 41: This needs to be an `AtomicBoolean` for thread-safety and should be passed as a dependency to all of the sub settings (Perhaps through a constructor)

This should also be private (Or at least private set)

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildData.kt
```diff
@@ -1,20 +1,44 @@
...
+import kotlin.concurrent.withLock
+import kotlin.reflect.jvm.javaType
+import kotlin.reflect.typeOf
 
-class GuildData(@field:Id val guildId: Long) {
+class GuildData(@field:Id val guildId: Long): Cloneable {
+
+    @Transient
+    @kotlin.jvm.Transient
+    var lock: ReadWriteLock = ReentrantReadWriteLock()
```
binaryoverload, line 37: Should be private and a `val` instead of `var`

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildData.kt
```diff
@@ -109,5 +138,27 @@ class GuildData(@field:Id val guildId: Long) {
 
     //endregion
 
+    fun write(writer: Consumer<GuildData>) {
+        this.lock.writeLock().withLock {
+            val copy: GuildData = CascadeBot.getGSON().fromJson(CascadeBot.getGSON().toJson(this), this.javaClass)
```
binaryoverload, line 143: Why... If we want to clone we should make a clone function of our own and not rely on GSON to do it properly!!

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildSettingsCore.kt
```diff
@@ -36,6 +37,8 @@ class GuildSettingsCore {
     @Setting
     var helpShowAllModules = false
 
+    var writeMode = false
```
binaryoverload, line 40: This should not be a separate field on its own, should use the main guild-data's one

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildSettingsManagement.kt
```diff
@@ -21,6 +21,8 @@ class GuildSettingsManagement {
     val greetings = Greetings()
     val autoRoles: MutableSet<Long> = mutableSetOf()
 
+    var writeMode = false
```
binaryoverload, line 24: See previous comment on these fields

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildSettingsCore.kt
```diff
@@ -105,10 +114,12 @@ class GuildSettingsCore {
     }
 
     fun addAlias(command: MainCommand, alias: String): Boolean {
+        if (!writeMode) throw UnsupportedOperationException("Cannot modify Guild data if not in write mode!")
```
binaryoverload, line 117: We should probably make a util function to do this, it's a lot of repeated code. `assertWriteMode()` perhaps?

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildSettingsModeration.kt
```diff
@@ -26,6 +26,8 @@ import java.util.function.Consumer
 @SettingsContainer(module = Module.MODERATION)
 class GuildSettingsModeration {
 
+    var writeMode = false
```
binaryoverload, line 29: See previous comment on these fields

src/main/java/org/cascadebot/cascadebot/utils/diff/Diff.java
```diff
@@ -0,0 +1,4 @@
+package org.cascadebot.cascadebot.utils.diff;
+
+public interface Diff {
```
binaryoverload, line 3: Instead of an empty interface, could have `getOld()` and `getNew()` functions?

binaryoverload (changes requested):

src/main/java/org/cascadebot/cascadebot/CascadeBot.java
weeryan17: Good question, it is a foreach which is usually blocking, but we're using the async client (and maybe the not async also is no blocking), I'll have to test this.

src/main/java/org/cascadebot/cascadebot/utils/diff/DiffUtils.java
weeryan17: Because it's easier, also I'm running into some reflection issues where java isn't finding all the fields, might have something to do with kotlin interopt. Might look into switching over the reflection in the future.

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildData.kt
weeryan17: The Transient annotation we're using is for polymorphia, the one I added is basically like adding transient to the variable in java

binaryoverload: Okay. Why do we need this?

weeryan17: So the copy of the guild data doesn't this stuff, so we don't diff it.

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildData.kt
weeryan17: https://github.com/CascadeBot/CascadeBot/pull/254#discussion_r642064489

src/main/java/org/cascadebot/cascadebot/utils/diff/Diff.java
weeryan17: because the things this is being used for don't all have old a new, this is also being used now in our CollectionDiff stuffs which doesn't have old or new, this interface is here so I have a common thing i can use when I'm referring to either the collection diff, or the old/new diff.

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildData.kt
weeryan17: (Although I just realized the new guild data won't end up getting the buttons and stuff copied over, so I'm going to need to account for that)

src/main/java/org/cascadebot/cascadebot/utils/diff/DiffUtils.java
weeryan17: Although switching over to reflection would resolve this https://github.com/CascadeBot/CascadeBot/pull/254#discussion_r642065697

weeryan17:
@binaryoverload when you get a chance this should(tm) be ready to be merged, granted we need to wait for modlog

weeryan17:
@binaryoverload @mrjvs if we could get this reviewed at some point that would be nice

src/main/java/org/cascadebot/cascadebot/data/managers/GuildDataManager.java
```diff
@@ -42,7 +55,36 @@
     public static void update(long id, Bson update) {
         CascadeBot.INS.getDatabaseManager().runAsyncTask(database -> {
             database.getCollection(COLLECTION, GuildData.class).updateOne(eq("_id", id), update, new DebugLogCallback<>("Updated Guild ID " + id + ":", update));
-        });
+        });;
```
binaryoverload: Extra semi colon :O

src/main/java/org/cascadebot/cascadebot/events/CommandListener.java
```diff
@@ -64,6 +64,7 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
         MDC.put("cascade.sender", event.getAuthor().toString());
         MDC.put("cascade.shard_info", event.getJDA().getShardInfo().getShardString());
         MDC.put("cascade.channel", event.getChannel().toString());
+        MDC.put("writeMode", String.valueOf(false));
```
binaryoverload: We don't need this anymore, it's on the guild data

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildSettingsManagement.kt
```diff
@@ -15,9 +17,16 @@ class GuildSettingsManagement {
...
+    
+
     val tags: ConcurrentHashMap<String, Tag> = ConcurrentHashMap()
     val filters: MutableList<CommandFilter> = mutableListOf()
     val permissions = GuildPermissions()
+    /*    get() {
+            assertWriteMode()
+            return field
+        }*/
+
```
binaryoverload: Need to look at why we did this

binaryoverload (changes requested):

src/main/kotlin/org/cascadebot/cascadebot/data/objects/GuildSettingsManagement.kt
weeryan17: I know why I did this, this was before we had the global write mode and permissions has a lot of writing so I just did it on everything, but then we added the global write mode and I added the assertWriteMode to the individual things so this is no longer needed.

binaryoverload (approved):

PR #258, opened 2021-06-04
Guild Data

binaryoverload:
- Introduce "inline write" to avoid having to wrap huge blocks of text
 - Instead of using MDC, use our own threadlocal on the GuildData companion object


weeryan17 (approved):
You beat me to it, I was going to do this stuff this weekend

